PR #8, opened 2019-12-20
readprop - allow text strings for object-type and property arguments.

ralight:
I'm not proposing this as a patch to be merged straight away. This change is an example of what I'd like to do across a few of the demo apps. If you like the idea, do you want me to duplicate the StrToX functions across the different apps, or to put them in a common file - and if so, then where would you suggest?

The existing method of calling bacrp requires the user to know the
enumeration value for the object-type and property values. This patch
enhances to allow the object-type and property arguments to be specified
as strings, using the strings as defined in the spec. It does not remove
the old behaviour.

Current:
  bacrp 1234 3 1 85

New:
  bacrp 1234 binary-input 1 present-value

### DIFF
--- a/apps/readprop/main.c
+++ b/apps/readprop/main.c
@@ -197,19 +197,23 @@ static void print_help(char *filename)
            "I-Am services.  For example, if you were reading\n"
            "Device Object 123, the device-instance would be 123.\n"
            "\nobject-type:\n"
-           "The object type is the integer value of the enumeration\n"
-           "BACNET_OBJECT_TYPE in bacenum.h.  It is the object\n"
-           "that you are reading.  For example if you were\n"
-           "reading Analog Output 2, the object-type would be 1.\n"
+           "The object type is object that you are reading. It\n"
+           "can be defined either as the object-type name string\n"
+           "as defined in the BACnet specification, or as the\n"
+           "integer value of the enumeration BACNET_OBJECT_TYPE\n"
+           "in bacenum.h. For example if you were reading Analog\n"
+           "Output 2, the object-type would be analog-output or 1.\n"
            "\nobject-instance:\n"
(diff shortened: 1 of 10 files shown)


skarg:
I like the idea of allowing text names in the demo apps! 

Keep in mind for function to have their return data types aligned, returning `long` to match strtol() rather than returning `int`.

I wonder if the functions should be centralized (perhaps in bactext.c)?  Something like:
`bool bactext_object_type_strtol(const char *search_name, unsigned *found_index)`
or
`long bactext_object_type_strtol(const char *search_name)`

ralight:
You're quite right about the return type thanks, I wasn't paying enough attention there.

I think that `bool bactext_object_type_strtol(const char *search_name, unsigned *found_index)` is the best choice, as long as you are happy with the change in behaviour for text strings that are neither an object type nor a number.

ralight:
This is now updated to include the functions as per your suggestion, and updated apps: readprop, readpropm (only object-type support, no property support), readrange, scov, ucov, whohas, writeprop, writepropm (only object-type support, no property support).

ralight:
Yes, that's a good idea. How about going even further and doing this:

```
static bool backtext_strtol_index(INDTEXT_DATA *istring, const char *search_name, unsigned *found_index)
{
    if(indtext_by_istring(istring, search_name, found_index) == true){
        return true;
    }else{
        value = strtol(search_name, &endptr, 0);
        if (endptr == search_name) {
        ... as in the current commit
    }
}

bool bactext_object_type_strtol(const char *search_name, unsigned *found_index)
{
    return backtext_strtol_index(bacnet_object_type_names, search_name, found_index);
}
```

ralight:
I've updated the commit with my latest suggestion. If you'd still prefer your way, just let me know.

PR #14, opened 2020-01-08
Make all demo object variables static.

ralight:
This conflicts with #12, but also solves the same issue. If you prefer to merge #12 first, then I'm happy to update this PR to fix the conflict.

### DIFF
--- a/src/bacnet/basic/object/ai.c
+++ b/src/bacnet/basic/object/ai.c
@@ -45,7 +45,7 @@
 #define MAX_ANALOG_INPUTS 4
 #endif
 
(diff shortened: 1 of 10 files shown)


ralight:
That's fixed now.

GeertL:
What about `Object_List` in `netport.c`? That might be another candidate to make static.

skarg:
The Travis build script was updated and caused check failures. You should be able to `git merge origin master` into your branch and push new changes to fix the checks for this pull request.

ralight:
@GeertL thanks, well spotted.

I've added the final static variable and rebased against the latest master, so this should be ok again.

PR #16, opened 2020-01-11
build: improve cmake, add github actions

gocarlos:
- compiles on windows
- adds gitlhub actions which execute CI for windows and linux (macos disabled as it is still failing)


gocarlos:
@skarg would be ready for review

PR #22, opened 2020-01-15
Fix gateway app build

skarg:


skarg:
This patch fixes the gateway build, which was broken and caused the new build process to block the merge.  The new build process checks are now blocked by lint errors. The lint errors can be fixed in a later patch.

PR #32, opened 2020-01-16
Update object property lists for all supported objects.

GeertL:
The property lists returned by `property_list_optional()` and `property_list_required()` were fairly outdated. This adds support for all objects currently in the library enums, and adds missing properties to the existing objects.

Some things I noticed, but did not change at the moment in order to avoid breaking existing code:

* Supported_Security_Algorithms property for Network Security object uses `PROP_SUPPORTED_SECURITY_ALGORITHM` enum, as well as `"supported-security-algorithm"` string; both are missing the `s` at the end.
* Trend Log object uses `OBJECT_TRENDLOG` rather than `OBJECT_TREND_LOG`. Note that Trend Log Multiple object _does_ use `OBJECT_TREND_LOG_MULTIPLE`.


skarg:
At the moment, master branch is targeting a 1.0.0 release, and in free form (all changes into master branch), so fixing or changing existing names or API is okay today.  After we tag 1.0.0 in master, then we can move active development into develop branch, apply bug fixes only to release branches where no API or name changes should occur.

PR #51, opened 2020-02-18
ci: add mac os to github ci

gocarlos:
this PR makes master pipeline green again... it became broken after activation of appveyor without the config file in the repository -> appveyor did not know what to do

fixing the share library build can be tackled in a later PR

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,10 @@
+configuration: Release
+
+image:
+  - Visual Studio 2019
+  - Ubuntu
+  - macOS
+
+build_script:
+  - cmake .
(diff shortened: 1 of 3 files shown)


gocarlos:
@madebr copied/adapted your appveyor file to make master green again...

madebr:
@gocarlos Cool!

Imho, you should still test `BUILD_SHARED_LIBS=ON` because making tests green should not be the goal.
The goal should be to be (as much as possible) error-free.
Even if it means that the tests are red..

gocarlos:
agree, this must be fixed, though upstream has to decide how to proceed with openssl -> in the meantime the pipeline should prevent to break other things...

PR #52, opened 2020-02-28
feat remove openssl dependency functions and add missing file to cmake

gocarlos:


src/bacnet/datalink/bacsec.h
```diff
@@ -211,10 +211,10 @@ extern "C" {
...
         uint32_t msg_len,
         uint8_t * signature);
-    bool key_verify_sign_msg(BACNET_KEY_ENTRY * key,
-        uint8_t * msg,
-        uint32_t msg_len,
-        uint8_t * signature);
+    // bool key_verify_sign_msg(BACNET_KEY_ENTRY * key,
+    //     uint8_t * msg,
+    //     uint32_t msg_len,
+    //     uint8_t * signature);
```
gocarlos: @skarg are you ok with removing those functions from the library? those functions call openssl functions which are not supported anymore and with bacnet secure probably also outdated?

skarg: Yes, I'm okay with removing the functions that require openssl dependency from the library.

Issue #17, opened 2020-01-12
remove visual studio project files, change to cmake as default build system on all platforms


gocarlos:
I guess there is no need to have `visual studio project files` anymore as visual studio works out of the box with cmake.

As cmake works on all platforms, its well supported by IDEs and recently also by package managers, I also suggest to make the default build system if possible.

PR #53, opened 2020-03-02
build: drop old visual studio / code blocks build files

gocarlos:
those files are not needed anymore as now we have cmake and visual studio / code blocks has support for cmake

furthermore those files contain the old project structure and are broken nonetheless...

fixes: https://github.com/bacnet-stack/bacnet-stack/issues/17


gocarlos:
@skarg what do you think?

PR #54, opened 2020-03-04
Export symbols in order to support shared libraries

### LINKED ISSUE #49: Building shared bacnet-stack libraries using cmake fails + openssl
Hello,

While helping to package this project at https://github.com/conan-io/conan-center-index/pull/909, I noticed some problems with bacnet-stack.

- When building shared libraries using cmake, the build fails.
I think this is because the linux build has not all files included.
- Also, some files use `OpenSSL`, but the cmake script does not look for it.
It requires a `find_package(OpenSSL)` somewhere + use of the variables/targets.
- macos fails to build

Fixing these issues should also fix some potential problems with the static build.

travis output
https://travis-ci.com/madebr/bacnet-stack/jobs/288603211
changes to `.travis.yml`:
https://github.com/madebr/bacnet-stack/commit/42041220cf1dc864614beadfa1be645b28cde550

madebr:
Fixes #49

This pr does the following:
- fix warnings bacause the definition `BAC_ROUTING` was not passed
- reorder cmake a bit in order to use less variables and do all port specific work in one place.
- Support both openssl 1.0 and 1.1 
  * I didn't see encryption support for mac and win, so I used the openssl code for mac and win also.
- When building a shared library, hide all symbols by default (using `-fvisibility=hidden`). This is the same behavior as MSVC. A symbol can be exported by prepending it with `BACNET_STACK_EXPORT`.
  I did not use the [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) helper as described [here](https://github.com/bacnet-stack/bacnet-stack/issues/49#issuecomment-592510005), because otherwise the custom `Makefile` build system would not work anymore.
- I tested mingw building on Linux

Problem(s) remaining:
- there now exists a hard dependency on OpenSSL. 
The CI fails because cmake cannot find openssl.
Github actions fails on macos and windows. Appveyor fails on on macos.
  * The github macos virtual environment has openssl installed (see [here](https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md), but I do not know where).
  * @gocarlos, you wrote the github actions ci. Do you know how to fix this?
- the cmake scripts should also be able to build all the apps/tests/targets that the Makefile system can build. But that is work for future pr's.


gocarlos:
oho, great :D
I'll have a look on it later

Great work!

gocarlos:
We can remove the openssl dependency from cmake

madebr:
I can also make the openssl dependency optional and put the `bacsec_linux.c` into the src tree as `bacsec.c` or something similar.

gocarlos:
In #53 we commented out the code which needs openssl as this was for an old bacnet profile... code could be removed in a next MR...

Therefore we can also remove it from cmake... better maintability

gocarlos:
LGTM 👍

PR #94, opened 2020-05-28
chore: remove BACAPP_SNPRINTF_ENABLED macro

gocarlos:
currently there is not way of setting this manually (without changing something in the cmake/make).

As I do not see the usefulness if this macro, just removing it...

@skarg do you have any platform that do not support this? if so, I'd change the behavior to be able to hide this code instead of hide by default.

### DIFF
--- a/src/bacnet/bacapp.c
+++ b/src/bacnet/bacapp.c
@@ -1439,7 +1439,7 @@ bool bacapp_parse_application_data(BACNET_APPLICATION_TAG tag_number,
                     value->type.Boolean = false;
                 break;
 #endif
-#if defined(BACAPP_UNSINED)
+#if defined(BACAPP_UNSIGNED)
             case BACNET_APPLICATION_TAG_UNSIGNED_INT:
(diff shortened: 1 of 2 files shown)


src/bacnet/bacapp.c
```diff
@@ -1404,50 +1430,67 @@ bool bacapp_parse_application_data(BACNET_APPLICATION_TAG tag_number,
...
+#if defined(BACAPP_BOOLEAN)
             case BACNET_APPLICATION_TAG_BOOLEAN:
                 long_value = strtol(argv, NULL, 0);
                 if (long_value)
                     value->type.Boolean = true;
                 else
                     value->type.Boolean = false;
                 break;
+#endif
+#if defined(BACAPP_UNSINED)
```
skarg, line 1442: misspelled define:  `BACAPP_UNSINED` should be `BACAPP_UNSIGNED` and is causing failure in unit test.

gocarlos: thanks

PR #95, opened 2020-05-31
fix: linking with cpp code

gocarlos:
this seems to fix a linking issue in my code on windows

example: 
```cpp
Send_I_Am(&Handler_Transmit_Buffer[0]);
```

@madebr you made the windows exports, does this make sense for you?

### DIFF
--- a/src/bacnet/basic/tsm/tsm.h
+++ b/src/bacnet/basic/tsm/tsm.h
@@ -21,8 +21,8 @@
 * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 *********************************************************************/
-#ifndef TSM_H
-#define TSM_H
+#ifndef BACNET_BASIC_TSM_TSM_H
+#define BACNET_BASIC_TSM_TSM_H
 
 #include <stdbool.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


madebr:
Makes total sense.
The name of that variable got mangled by the c++ compiler.

src/bacnet/basic/tsm/tsm.h
```diff
@@ -21,8 +21,8 @@
 * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 *********************************************************************/
-#ifndef TSM_H
-#define TSM_H
+#ifndef BACNET_BASIC_TSM_TSM_H
+#define BACNET_BASIC_TSM_TSM_H
```
gocarlos, line 25: I think that we should start changing those guards when we touch a file as they can cause trouble with legacy apps e.g. 

config.h uses `CONFIG_H`, abort.h uses `ABORT_H` etc. which are probably used also by other old libraries

best would be imho the usage of `#pragma once` as then you never have copy/paste issues

madebr: If you use a private prefix, then people using that prefix will enter undefined behaviour.

gocarlos: why private prefix? this is just a namespaced prefix to not collide with other libraries...

basically the same as https://github.com/bacnet-stack/bacnet-stack/pull/91

madebr: I meant the same thing. Use a private namespace (= private prefix) in the include guard. (I don't mean c++ namespaces)

PR #99, opened 2020-06-22
Update README.md

sdghchj:
Add more instructions.

I am not clear about some env variables, so it still needs improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ c:\> cd build/
 c:\> cmake .. -G"CodeBlocks - MinGW Makefiles"
 ```
 
-##Runtime envs
+## Runtime envs
 
 common:
 
@@ -192,7 +192,7 @@ BACNET_MSTP_MAC:
 
 
 
-##Demos
+## Demos
 
 The demo applications are all client applications that provide one main BACnet
 service, except the one server application and one gateway application, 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,13 +140,59 @@ The library is also instrumented to use [CMake](https://cmake.org/) which can
...
+$ mkdir build
+$ cd build/
+$ cmake .. -G"CodeBlocks - Unix Makefiles"
+
+c:\> mkdir build
+c:\> cd build/
+c:\> cmake .. -G"CodeBlocks - MinGW Makefiles"
+```
+
+## Runtime envs
```
skarg, line 153: The runtime environment variables are shown in [bin/readme.txt](https://github.com/bacnet-stack/bacnet-stack/blob/master/bin/readme.txt) file. No need to duplicate them here.

README.md
```diff
@@ -126,6 +208,19 @@ and others.  Each of those projects has a demo application for specific hardware
 In the case of the ARM7 and AVR, their makefile works with GCC compilers and
 there are project files for IAR Embedded Workbench and Rowley Crossworks for ARM.
 
+Commonly used apps:
```
skarg, line 211: Please reference the [bin/readme.txt](https://github.com/bacnet-stack/bacnet-stack/tree/master/bin) which already has this information.

README.md
sdghchj: So, since you have already prepared these instructions, why not put them in the portal doc? One have to go deep into these directories to search for useage?

PR #162, opened 2021-04-02
Bugfix/gateway device application

skarg:


skarg:
mixed two branches - oops!

PR #228, opened 2022-02-22
Feature/MSTP promiscuous read

NickSchaf-jci:
This adds two callback functions to dlmstp (implemented on the stm32f4xx port at this time) so that the caller may observe incoming frames on the MSTP bus, regardless of whether they are addressed to the local MSTP address.

The callback functions, if provided, are called from `MSTP_Receive_Frame_FSM()` instead of `dlmstp_receive()` in order to:

- catch the preamble for noting the start of a frame
- call the frame completion callback as soon as the frame is complete
- not care whether the frame is for us or "NotForUs"

Sample Usage:
```
uint32_t frame_start_time = 0;

void mstp_rx_frame_start() {
    frame_start_time = get_ticks();
}

void mstp_rx_promiscuous_callback(uint8_t src, uint8_t dest, uint8_t mstp_msg_type, uint8_t * pdu, uint16_t pdu_len) {
    if (FRAME_TYPE_TOKEN == mstp_msg_type) {
        printf("%u handed token to %u. token frame took %u ms\r\n", src, dest, get_ticks() - frame_start_time);
    }
}

void app_main(void) {
    /* Other initialization, declarations, etc. here */

    dlmstp_init(NULL);
    dlmstp_set_frame_rx_start_callback(mstp_rx_frame_start);
    dlmstp_set_frame_rx_complete_callback(mstp_rx_promiscuous_callback);

    while (true) {
        pdu_len = dlmstp_receive(&src_addr, mstp_rx_buffer, DLMSTP_MPDU_MAX, 0);
    }
}
```

Additionally: when testing these changes, it was noted that when the `MSTP_Flag.ReceivedValidFrame` flag is set, `ReceiveFrameCount` could be incremented multiple times for the same frame.   This happens because `dlmstp_receive()` may be called several times between the frame arriving and `rs485_turnaround_elapsed()` returning true.
To fix this, `ReceiveFrameCount++` for "our" frames was moved to after the `rs485_turnaround_elapsed()` test.

### DIFF
--- a/ports/stm32f4xx/dlmstp.c
+++ b/ports/stm32f4xx/dlmstp.c
@@ -175,6 +175,12 @@ static unsigned ReceivePDUCount;
 static volatile struct dlmstp_packet PDU_Buffer[MSTP_PDU_PACKET_COUNT];
 static RING_BUFFER PDU_Queue;
 
+/* Callback function to be called every time we receive a preamble */
+static dlmstp_hook_frame_rx_start_cb PreamCallback = NULL;
+
+/* Callback function to be called every time we receive a frame */
+static dlmstp_hook_frame_rx_compl_cb FrameRxCallback = NULL;
+
 bool dlmstp_init(char *ifname)
 {
     ifname = ifname;
@@ -442,6 +448,11 @@ static void MSTP_Receive_Frame_FSM(void)
                     /* Preamble1 */
                     /* receive the remainder of the frame. */
                     Receive_State = MSTP_RECEIVE_STATE_PREAMBLE;
+
+                    /* if a frame-start callback was provided, call it */
+                    if (PreamCallback != NULL) {
+                      PreamCallback();
+                    }
                 }
             }
             break;
@@ -552,6 +563,10 @@ static void MSTP_Receive_Frame_FSM(void)
                                 /* NotForUs */
                                 MSTP_Flag.ReceivedValidFrameNotForUs = true;
                             }
+                            /* if a frame-receipt callback was provided, call it for this frame */
+                            if (FrameRxCallback != NULL) {
+                              FrameRxCallback(SourceAddress, DestinationAddress, FrameType, InputBuffer, DataLength);
+                            }
                             /* wait for the start of the next frame. */
                             Receive_State = MSTP_RECEIVE_STATE_IDLE;
                         } else {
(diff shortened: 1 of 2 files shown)


ports/stm32f4xx/dlmstp.c
```diff
@@ -175,6 +175,12 @@ static unsigned ReceivePDUCount;
 static volatile struct dlmstp_packet PDU_Buffer[MSTP_PDU_PACKET_COUNT];
 static RING_BUFFER PDU_Queue;
 
+/* Callback function to be called every time we receive a preamble */
+static dlmstp_hook_frame_rx_start_cb PreamCallback = NULL;
+
+/* Callback function to be called every time we receive a frame */
+static dlmstp_hook_frame_rx_compl_cb FrameRxCallback = NULL;
```
skarg, line 182: please rename `dlmstp_hook_frame_rx_compl_cb` to `dlmstp_hook_frame_rx_complete_cb`
Abbreviations may unintentionally encrypt the words and not be understandable by others.

ports/stm32f4xx/dlmstp.c
```diff
@@ -175,6 +175,12 @@ static unsigned ReceivePDUCount;
 static volatile struct dlmstp_packet PDU_Buffer[MSTP_PDU_PACKET_COUNT];
 static RING_BUFFER PDU_Queue;
 
+/* Callback function to be called every time we receive a preamble */
+static dlmstp_hook_frame_rx_start_cb PreamCallback = NULL;
```
skarg, line 179: Please rename `PreamCallback` to `Preamble_Callback`
Abbreviations may unintentionally encrypt the words and not be understandable by others.
Underscores between words improve readability. Yep, it's a double standards as the BACnet standard is full of words put together without underscores that are brought into this codebase as-is.

ports/stm32f4xx/dlmstp.c
```diff
@@ -442,6 +448,11 @@ static void MSTP_Receive_Frame_FSM(void)
                     /* Preamble1 */
                     /* receive the remainder of the frame. */
                     Receive_State = MSTP_RECEIVE_STATE_PREAMBLE;
+
+                    /* if a frame-start callback was provided, call it */
+                    if (PreamCallback != NULL) {
+                      PreamCallback();
```
skarg, line 454: please run [clang-format](https://llvm.org/builds/) on these code changes, or just the new code before submitting so that the style matches the existing style (in the root .clang-format file).

ports/stm32f4xx/dlmstp.c
NickSchaf-jci: Fixed

ports/stm32f4xx/dlmstp.c
NickSchaf-jci: Fixed this.  Also followed suit by renaming `FrameRxCallback` to `Frame_Rx_Callback`

ports/stm32f4xx/dlmstp.c
NickSchaf-jci: Updated

PR #230, opened 2022-02-24
Avoid redefining MSTP timeouts

NickSchaf-jci:
There was a precedent in src/bacnet/datalink/mstp.c to check whether timeouts were already defined before defining them.
Example:
```
#ifndef Treply_timeout
#define Treply_timeout 295
#endif
```

This change extends those `#ifndef` checks to the various ports so that timeouts may be overridden by defining them in bacnet-config.h (assuming `BACNET_CONFIG_H` is defined to cause bacnet-config.h to be included)

### DIFF
--- a/ports/at91sam7s/dlmstp.c
+++ b/ports/at91sam7s/dlmstp.c
@@ -119,12 +119,6 @@ static uint8_t Nmax_info_frames = 1;
 /* its value shall be 127. */
 static uint8_t Nmax_master = 127;
 
-/* The time without a DataAvailable or ReceiveError event before declaration */
-/* of loss of token: 500 milliseconds. */
-#ifndef Tno_token
-#define Tno_token 500
-#endif
-
 /* The minimum time without a DataAvailable or ReceiveError event */
 /* that a node must wait for a station to begin replying to a */
 /* confirmed request: 255 milliseconds. (Implementations may use */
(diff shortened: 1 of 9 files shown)


skarg (changes requested):
Some of the MSTP defines are fixed definitions, and shouldn't have the option of changing.  These fixed defines are defined in [mstpdef.h](https://github.com/bacnet-stack/bacnet-stack/blob/master/src/bacnet/datalink/mstpdef.h) file.

Can you include the mstpdef.h file in these files, and eliminate the duplicate usage of the fixed defines?

NickSchaf-jci:
@skarg Cleaned up the ports' MS/TP implementations by removing those time definitions that were already defined in mstpdef.h.  As it turns out, all those ports already included mstpdef.h (either directly or indirectly via [src/datalink/mstp.h](https://github.com/bacnet-stack/bacnet-stack/blob/master/src/bacnet/datalink/mstp.h)).

I left the other time definitions as-is (with the redefinition checks) in the ports.  There are several instances where they were using different values.  Example:
```
$ grep -rn --include="dlmstp.c" --include="mstp.c" "#define Treply_timeout" *
ports/at91sam7s/dlmstp.c:127:#define Treply_timeout 260
ports/atmega168/dlmstp.c:143:#define Treply_timeout 295
ports/bdk-atxx4-mstp/dlmstp.c:131:#define Treply_timeout 260
ports/pic18f6720/mstp.c:122:#define Treply_timeout 255
ports/pic18f97j60/mstp.c:122:#define Treply_timeout 255
ports/stm32f10x/dlmstp.c:118:#define Treply_timeout 260
ports/stm32f4xx/dlmstp.c:131:#define Treply_timeout 260
ports/xplained/dlmstp.c:131:#define Treply_timeout 260
src/bacnet/datalink/mstp.c:153:#define Treply_timeout 295
```

skarg (approved):

PR #232, opened 2022-02-27
Bugfix/clang tidy applied 2022

skarg:


apps/perl/perl_bindings.c
```diff
@@ -64,7 +64,7 @@ static void MyAbortHandler(
     if (address_match(&Target_Address, src) &&
         (invoke_id == Request_Invoke_ID)) {
         char msg[MAX_ERROR_STRING];
-        sprintf(msg, "BACnet Abort: %s",
+        sprintf(msg, "BACnet Abort: %d",
```
AlexanderWells-diamond, line 67: @skarg I think something has gone wrong here - the baxtext_* functions all quite clearly return a const char* ? Won't this end up just printing a pointer?

PR #242, opened 2022-03-21
basic/object/iv: Name it "INTEGER VALUE", not "ANALOG VALUE"

laumann:
Running the sample BACnet demo server, we noticed that the names of
integer-value types were all "ANALOG VALUE <n>" for some number n which
creates multiple objects with the same object-name.

Signed-off-by: Thomas Jespersen <tj@omnio.net>

---
Thanks for a great library btw, with well-working demos to build off of!


skarg (approved):

laumann:
@skarg Nice, thank you! Would you accept a PR to change the comments in `iv.c` to to "Integer value" instead of "Analog value"?

skarg:
Yes, corrected comments or documentation are always welcome.

PR #271, opened 2022-05-11
OSCBS-23 Scalable BACnet object arrays

Michail-Antropov:
This is sample of way to do OSCBS-23 
 - Copies src/basic/object/*.c files put to ports/zephyr folder.
 - Add STRUCT_SECTION_ITERABLE
 "Main" changed file is ports/zephyr/device.c

Tested on zephyr/samples/profiles/b-ss/
At least test is compiled and run.

### DIFF
--- a/ports/zephyr/acc.c
+++ b/ports/zephyr/acc.c
@@ -1,0 +1,453 @@
+/**************************************************************************
+ *
+ * Copyright (C) 2017 Steve Karg <skarg@users.sourceforge.net>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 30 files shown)


gregshue:
@Michail-Antropov We need to refactor this to minimize the logic and definitions beneath `ports/*/`. The OBJECT_TABLE design should be portable across multiple platforms/environments, so should be under `src/`. For this design, only the definition of `BACNET_OBJECT_TABLE` should be beneath `ports/*/`.

Since the mapping of object IDs to specific data is project-specific, this mapping should be under `$bacnet-stack/zephyr/` rather than under `src/` or `ports/zephyr/`.

Please refactor.

PR #281, opened 2022-05-25
Make "dlenv_network_port_init()" public (was static)

MightyPork:
This is needed when initing as a library, without calling "dlenv_init()". 

That's desirable if you don't want to use env variables and don't want to risk hitting exit() in the function (that should probably be removed).


MightyPork:
CI errors should now be resolved

PR #283, opened 2022-05-25
Add option to subscribe to single property COV

MightyPork:
This is experimental.

I wanted to receive just property 85 COV, but the server sends COV also for other properties of the object.

This was an attempt to send a more selective COV subscription. It did not work in my case, I still receive other properties.

Would be interesting to verify the function against some other bacnet server.


MightyPork:
Okay, so does this mean I will receive `Status_Flags` COV whenever I subscribe to `Present_Value`, even if I use the `SubscribeCOVProperty` service? So the utility of `SubscribeCOVProperty` is that you can subscribe to other properties than `Present_Value`?

skarg:
I reviewed the code and using cov_subscribe_property_encode_apdu() should subscribe to one property, and the COV server should only send a COV Notification with one property value, if it supports the service.

I'd be interested in seeing a Wireshark capture of the subscribe and subsequent notifications.

PR #288, opened 2022-05-28
Fix error in datalink/bvlc.c

Michail-Antropov:
1 In bvlc_foreign_device_bbmd_host_address_decode() remove checking of close the common tag.
Reason - the bvlc_foreign_device_bbmd_host_address_encode() is not encode common tag for BACnetHostNPort and bvlc_foreign_device_bbmd_host_address_decode() is not check open tag.
Add and check tag for host only.
Test - test/bacnet/datalink/bvlc.

### DIFF
--- a/test/bacnet/datalink/bvlc/src/main.c
+++ b/test/bacnet/datalink/bvlc/src/main.c
@@ -820,7 +820,7 @@ static void test_BVLC_BBMD_Address(void)
 {
     uint8_t apdu[480] = { 0 };
     uint16_t apdu_len = 0;
-    uint16_t test_apdu_len = 0;
+    int16_t test_apdu_len = 0;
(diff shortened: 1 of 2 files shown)


test/bacnet/datalink/bvlc/src/main.c
```diff
@@ -820,7 +820,7 @@ static void test_BVLC_BBMD_Address(void)
 {
     uint8_t apdu[480] = { 0 };
     uint16_t apdu_len = 0;
-    uint16_t test_apdu_len = 0;
+    int16_t test_apdu_len = 0;
```
Michail-Antropov, line 823: bvlc_foreign_device_bbmd_host_address_decode returns signed value (negative value for errors)

skarg (approved):

PR #293, opened 2022-06-10
BIP: add option - listen to broadcast on a separate socket

Michail-Antropov:
It allows start plural BACnet apps on single computer.

How to
1 Create lo:X loopbacks:
`sudo ifconfig lo:1 127.0.0.2 netmask 255.0.0.0 up`
`sudo ifconfig lo:2 127.0.0.3 netmask 255.0.0.0 up`
`sudo ifconfig lo:3 127.0.0.4 netmask 255.0.0.0 up`

2 Build a BACnet application with the "Use a separate socket for listening broadcast messages" option
`~/bacnetplatform/bacnet-stack/apps$ make clean`
`~/bacnetplatform/bacnet-stack/apps$ make BACDL_BROADCAST=1 server`

3 Run the app one on each loopback:
`BACNET_IFACE="lo:1" ./bacserv 260002`
`BACNET_IFACE="lo:2" ./bacserv 260003`

4 Run another BACnet utility on another loopback (if necessary):
Ex. Read Property from apps/readprop
`BACNET_IFACE="lo:3" ./bacrp --mac 127.0.0.2:47808 2600002 18 260002 121`


cybermag:
Additional work should be done on that pull request: if BACDL_BROADCAST=1, the code must prevent duplication of broadcast packets.

skarg:
I think the separate BACnet/IP broadcast will be a required feature, so no need to make it compile optional.  See:
https://sourceforge.net/p/bacnet/bugs/65/

In addition to having bvlc_handler(), we need to add bvlc_broadcast_handler() so that BVLC_ORIGINAL_UNICAST_NPDU can be dropped when sent as broadcast.

We should also do the same (add additional broadcast socket) for BSD/MacOS and Windows for BACnet/IP.

cybermag:
@skarg Are you sure that we want to make refactoring of fixes related to https://sourceforge.net/p/bacnet/bugs/65/ ? Basically all what we are going to do is to add one more socket (which can receive broadcast packets ony) into descriptor array for select in bip_receive() and bip6_receive() function. Such additional broadcast socket is going to be added only in a case if interface does not have SO_BROADCAST flag. So all code will work as it works without that fix. But I don't understand your idea about  bvlc_handler() and  bvlc_broadcast_handler(). The mentioned fix is in https://sourceforge.net/p/bacnet/bugs/65/ so it should be moved to some other place. Can you please to clarify your idea more detaily?

cybermag:
@skarg Thanks, that's help a lot. But what about fix related to https://sourceforge.net/p/bacnet/bugs/65?
(hack for 5.4.5.1). 
I can see 3 places with that fix:
image
Does it mean that code which drops broadcasts in that code shall be removed?

skarg:
There are 2 types of broadcast indications:
1. NPDU header dnet=65535 indicates broadcast packet (hack for 5.4.5.1 IDLE)
2. datalink broadcast frame (specific to each datalink)

BACnet/IP in the stack doesn't have a mechanism to determine or handle the datalink broadcast frame at the moment.

Ports like ports/bsd and ports/windows have a bug if bip_init() doesn't call bvlc_init()

ports/linux/bip-init.c
```diff
@@ -508,6 +525,28 @@ int bip_get_local_address_ioctl(char *ifname, struct in_addr *addr, int request)
...
+ * @brief Issue a specific request foor an interface via an ioctl() call.
+ * @param ifname - the interface name
+ * @param flags [out] the active flag word of the device.
+ * @param request - the ioctl() request
+ * @return 0 on success, else the error from the ioctl() call.
+ */
+int bip_get_flags_ioctl(char *ifname, short *flags, int request)
+{
+    struct ifreq ifr = { { { 0 } }, { { 0 } } };
+    struct sockaddr_in *tcpip_address;
```
gregshue, line 538: Where is this used?  If not, please remove it.

gregshue (changes requested):

skarg (approved):

ports/linux/bip-init.c
Michail-Antropov: Fix

BACnetEd:
Gents, 

I don't think this is the right approach.

- It only addresses the UDP datalink, and the exact same issue arises in MS/TP
- I don't see how listening on e.g. 192.168.1.11 and simultaneously 192.168.1.255 disambiguates unicast from broadcast. I have not experimented, but I have asked StackOverflow (and there is a clue there already) : https://stackoverflow.com/questions/77253483/how-does-one-detect-upon-receipt-that-a-udp-packet-was-sent-using-a-broadcas - but, again, this does not help with MS/TP
- Note, the "Ignore Confirmed messages"  in 5.4.5.1 applies to the Application Entity (AE), not the Network Entity (NE), so we should not even be filtering at the BVLL or Network Layers
- In fact, there are tests 10.7.4 and 10.7.5 that make sure the NE passes on broadcast messages.

As for other approaches I have seen:

- Nothing at the BVLL, should be looking at DER (Which is a Network Layer construct)
- Nothing at the BVLL or the Network Layer should be looking at the Application Layer (i.e. Detecting "ConfirmedPDU")
- I have seen a MSG_PEEK function that I fear may not be portable
- But these are, IMHO, academic - see below

I believe the correct approach should be simply to:

- Look at the BVLC function, detect "Original-Broadcast-NPDU" or "Distribute-Broadcast-To-Network"
- And PASS THIS INFORMATION up the stack, (i.e. initially to the Network Layer), using N-UNITDATA.indication
- How? (There is no 'broadcast flag' in the N-UNITDATA.indication). So use the destination_address (MAC, len=0, indicates broadcast)
- Only when this payload (APDU) along with "metadata" (broadcast "flag"), reaches the AE, should the AE examine the type of message, and if confirmed and broadcast, simply ignore it.

PR #296, opened 2022-06-13
Added Kconfig options to keep small footprint for MCUs having less RAM

honestech74:
For now, Zephyr build will occur out of SRAM error with boards having RAM less than about 1MB.
This added options to enable/disable the basic objects including command, schedule and trendlog. It uses about 40kb of memory for bacnet-stack with the following definitions.
```
  MULTISTATE_NUMBER_OF_STATES=3
  MAX_TSM_TRANSACTIONS=3
```
Also it will build with `zephyr/samples/profile/b-ss` on `nrf52840dk_nrf52840` board.

### DIFF
--- a/ports/zephyr/device.c
+++ b/ports/zephyr/device.c
@@ -1,0 +1,1898 @@
+/**************************************************************************
+ *
+ * Copyright (C) 2005,2006,2009 Steve Karg <skarg@users.sourceforge.net>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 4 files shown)


gregshue (changes requested):
Thank you for the proposal. We have other work going on to make `device.c` scalable.  This is a different implementation built on Zephyr's iterable arrays. Let's leave this one pending while we align on how that architectural variation is organized.  Those PRs are available for review if you want to track the discussion.

skarg:
Utilized this concept and updated a) bacnet-config.h for zephr OS with low RAM defines, and b) Kconfig around every BACnet basic object file to enable inclusion or not when build the library.  See [Bugfix/zephyr build for minimal ram](https://github.com/bacnet-stack/bacnet-stack/pull/606) for specifics.

PR #310, opened 2022-07-25
OSCBS-33 add pipeline script

Michail-Antropov:


skarg:
This bacnet-stack open source project runs github actions, not bitbucket CI. Could you modify the testing script to use github actions using .github/workflows/zephyr.yml file? I don't think there is a need for artifacts at the moment.

Zephyr RTOS creates their own docker image:
https://github.com/zephyrproject-rtos/docker-image
Puts it here:
https://github.com/zephyrproject-rtos/docker-image/pkgs/container/ci
Uses it from their github actions:
https://github.com/zephyrproject-rtos/zephyr/blob/main/.github/workflows/twister.yaml

Rather than use their docker image directly (with 'latest' tag):
`docker pull ghcr.io/zephyrproject-rtos/ci:latest`
use from [docker hub:](https://hub.docker.com/r/zephyrprojectrtos/zephyr-build/tags):
`docker pull zephyrprojectrtos/zephyr-build:latest`

Michail-Antropov:
Does not actual

PR #319, opened 2022-08-02
Schedule encoding/decoding

MightyPork:
This is a work in progress, feel free to comment on it to guide the implementation.

Currently it compiles, hasn't been tested yet. It also doesn't have unit tests for schedule.

TimeValue had to be changed to use `BACNET_SHORT_APPLICATION_DATA_VALUE` - it would not compile when I used the normal data value struct.

I also had to rename the struct in object/schedule.c due to name conflict.


MightyPork:
No idea what went wrong in the windows build this time

skarg:
I'd prefer to rename BACNET_SHORT_APPLICATION_DATA_VALUE to something other than "SHORT".  The BACnetTimeValue is described in the BACnet standard as:
```
BACnetTimeValue ::= SEQUENCE {
    time Time,
    value ABSTRACT-SYNTAX.&Type -- any primitive datatype; complex types cannot be decoded
}
```
Perhaps BACNET_PRIMATIVE_VALUE, although it is really just limiting to 32-bit values (except on compiler that allow 64-bit integers).

It's too bad that the values of BACnetPriorityValue are not constrained enough or that would be a good definition and name:
```
BACnetPriorityValue ::= CHOICE {
    null NULL,
    real REAL,
    enumerated ENUMERATED,
    unsigned Unsigned,
    boolean BOOLEAN,
    integer INTEGER,
    double Double,
    time Time,
    characterstring CharacterString,
    octetstring OCTET STRING,
    bitstring BIT STRING,
    date Date,
    objectidentifier BACnetObjectIdentifier,
    constructed-value [0] ABSTRACT-SYNTAX.&Type,
    datetime [1] BACnetDateTime
}
```
Also, please don't do magic "up-casting" to BACNET_APPLICATION_DATA_VALUE - structure members aren't guaranteed exact binary placement.  Remove dummy structure members, and add pre/post copy as needed, or create separate handling functions for the BACNET_PRIMATIVE_VALUE data structure values and include BACNET_PRIMATIVE_VALUE directly in BACnet_Application_Data_Value structure.

MightyPork:
I don't see how the two structs could be aligned differently, but it's a relatively small change so why not, it's done. There's some overhead incurred in the conversion, though. I also renamed the struct to BACNET_PRIMITIVE_DATA_VALUE.

I had to change the schedule object implementation in schedule.c/.h so `Present_Value` is now a copy instead of a pointer into the struct, this was the place where casting was really needed.

skarg:
Please run clang-format on the changed files so that the style will be aligned.

MightyPork:
Formatted, sorry it takes so long, we're using the fork and it works fine, so this got deprioritized. 

Now there's a conflict, not sure what to do with that. It'd be easier if you resolved it when merging... or I can try to rebase.

MightyPork:
note that clang-format changed a lot of things I never touched, so the diff is now huge ... I think this also caused the conflict.

skarg:
Yikes! Sorry about that. If you want to revert the clang-format I can do bulk format after the merge.

MightyPork:
it's now force-pushed to the commit before formatting

skarg (approved):

PR #337, opened 2022-09-20
RPM should return UNKNOWN_OBJECT for non-existent objects

rcb-tcs:
Right now, the stack is returning empty values/lists, which I understand is not in accordance with the spec.

### DIFF
--- a/src/bacnet/basic/service/h_rpm.c
+++ b/src/bacnet/basic/service/h_rpm.c
@@ -236,6 +236,17 @@ void handler_read_property_multiple(uint8_t *service_request,
                     berror = true;
                     break;
                 }
+                
+                /* Test for invalid object before proceeding */
+                if (!Device_Valid_Object_Id(rpmdata.object_type, rpmdata.object_instance)) {
+                    rpmdata.error_class = ERROR_CLASS_OBJECT;
+                    rpmdata.error_code =
+                        ERROR_CODE_UNKNOWN_OBJECT;
+                    error = BACNET_STATUS_ERROR;
+                    berror = true;
+                    break; /* The berror flag ensures that both loops will */
+                           /* be broken! */
+                }
 
                 /* Test for case of indefinite Device object instance */
                 if ((rpmdata.object_type == OBJECT_DEVICE) &&
--- a/src/bacnet/basic/service/h_rpm.c
+++ b/src/bacnet/basic/service/h_rpm.c
@@ -236,17 +236,6 @@ void handler_read_property_multiple(uint8_t *service_request,
                     berror = true;
                     break;
                 }
-                
-                /* Test for invalid object before proceeding */
-                if (!Device_Valid_Object_Id(rpmdata.object_type, rpmdata.object_instance)) {
-                    rpmdata.error_class = ERROR_CLASS_OBJECT;
-                    rpmdata.error_code =
-                        ERROR_CODE_UNKNOWN_OBJECT;
-                    error = BACNET_STATUS_ERROR;
-                    berror = true;
-                    break; /* The berror flag ensures that both loops will */
-                           /* be broken! */
-                }
 
                 /* Test for case of indefinite Device object instance */
                 if ((rpmdata.object_type == OBJECT_DEVICE) &&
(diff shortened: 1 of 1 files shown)


skarg:
In the existing implementation, ReadProperty and ReadPropertyMultiple handlers rely on Device_Read_Property() to return an error if the object is not found:
```int Device_Read_Property(BACNET_READ_PROPERTY_DATA *rpdata)
{
    int apdu_len = BACNET_STATUS_ERROR;
    struct object_functions *pObject = NULL;

    /* initialize the default return values */
    rpdata->error_class = ERROR_CLASS_OBJECT;
    rpdata->error_code = ERROR_CODE_UNKNOWN_OBJECT;
    pObject = Device_Objects_Find_Functions(rpdata->object_type);
    if (pObject != NULL) {
        if (pObject->Object_Valid_Instance &&
            pObject->Object_Valid_Instance(rpdata->object_instance)) {
            apdu_len = Read_Property_Common(pObject, rpdata);
        } else {
            rpdata->error_class = ERROR_CLASS_OBJECT;
            rpdata->error_code = ERROR_CODE_UNKNOWN_OBJECT;
        }
    } else {
        rpdata->error_class = ERROR_CLASS_OBJECT;
        rpdata->error_code = ERROR_CODE_UNKNOWN_OBJECT;
    }

    return apdu_len;
}
```
Does your device object include this kind of code?

rcb-tcs:
I just now checked, and it does look like for an invalid object we're effectively exiting ```Device_Read_Property``` with:

    rpdata->error_class = ERROR_CLASS_OBJECT;
    rpdata->error_code = ERROR_CODE_UNKNOWN_OBJECT;
    return BACNET_STATUS_ERROR;

I didn't mention, but I was mostly testing with the all/required/optional special properties, and for those ```Device_Read_Property``` never gets called, just ```Device_Objects_Property_List```. For an unknown object, at least in our implementation, we just return a set of empty ```special_property_lists``` invalid and dump the error since that function can't return one. So when I suggested this change, that was the first thing on my mind.

Now that I'm looking a little more at ```Device_Read_Property```, I noticed there does still seems to be something funny with RPM for regular properties of unknown objects too. If I do an RPM on, let's say, 3 different regular properties of a unknown object, what I get is a Complex-ACK response with 3 propertyAccessErrors encapsulating the UNKNOWN_OBJECT errors returned by ```Device_Read_Property```.

skarg:
It seems that the special case is the one with `property_count == 0` AND an unknown object.  It probably should be solved in that case alone (not by aborting the whole RPM). Perhaps the code change should be (added to existing special case):
```
                            if (property_count == 0) {
                                /* This only happens with the OPTIONAL property
                                 */
                                /* 135-2016bl-2. Clarify ReadPropertyMultiple
                                   response on OPTIONAL when empty. */
                                /* If no optional properties are supported then
                                   an empty 'List of Results' shall be returned
                                   for the specified property, except if the
                                   object does not exist. */
                                if (!Device_Valid_Object_Id(rpmdata.object_type,
                                  rpmdata.object_instance)) {
                                    len = RPM_Encode_Property(
                                        &Handler_Transmit_Buffer[npdu_len],
                                        (uint16_t)apdu_len, MAX_APDU, &rpmdata);
                                    if (len > 0) {
                                        apdu_len += len;
                                    } else {
#if PRINT_ENABLED
                                        fprintf(stderr,
                                            "RPM: Too full for property!\r\n");
#endif
                                        error = len;
                                        /* The berror flag ensures that
                                           both loops will be broken! */
                                        berror = true;
                                        break;
                                    }
                                }
```

rcb-tcs:
Ah I see. That makes sense and seems to be the only change needed. I'll change the pull request to reflect that.

It makes sense that the UNKNOWN_OBJECT error should be encapsulated in the Complex-ACK response whenever it occurs. Before, I had thought only one object could be accessed by RPM at a time, which is why I expected an unknown object ought to fail the whole request.

skarg (approved):

skarg:
Since the BACnet standard allows returning the special RPM properties as part of the results in errors, we should move the unknown object for OPTIONAL to include ALL, REQUIRED, OPTIONAL.  Could you change your pull request:
1. move the object exist check to an `if ()` clause just before `if (rpmdata.array_index != BACNET_ARRAY_ALL) {` and make the `array_index` check into an `else if {}` clause?
2. add a bypass of the `Device_Read_Property()` function for the special properties in `RPM_Encode_Property()` function:
```
    if ((rpmdata->object_property == PROP_ALL) ||
        (rpmdata->object_property == PROP_REQUIRED) ||
        (rpmdata->object_property == PROP_OPTIONAL)) {
        /* special properties only get ERROR encoding */
        len = BACNET_STATUS_ERROR;
    } else {
        len = Device_Read_Property(&rpdata);
    }
```

rcb-tcs:
Thanks for the feedback! I did go ahead and make/push those changes

PR #348, opened 2022-10-13
Fix decoding of sentConfirmedNotifications request

stevedh:
Description: confirmed COV notifications are generated even when the subscription included the `issueConfirmedNotificationsFlag=false` flag.

`issueConfirmedNotificationsFlag` is context tag 2, optional; so it could be that most implementations will omit the field rather than including it and setting it to false.

I *think* the decoding logic is wrong since boolean values are encoded using the length field, so this seems like the right fix.

Here's a Wireshark screenshot of a packet which seems confirming; after receiving this message bacnet-stack will generate confirmed notifications (in this trace I have applied the patch so you see the correct behavior).

Screen Shot 2022-10-13 at 10 04 00 AM

### DIFF
--- a/src/bacnet/cov.c
+++ b/src/bacnet/cov.c
@@ -489,7 +489,7 @@ int cov_subscribe_decode_service_request(
                 data->cancellationRequest = false;
                 len += decode_tag_number_and_value(
                     &apdu[len], &tag_number, &len_value);
-                data->issueConfirmedNotifications = len_value;
+                data->issueConfirmedNotifications = decode_boolean(len_value);
             } else {
                 data->cancellationRequest = true;
             }
@@ -652,7 +652,7 @@ int cov_subscribe_property_decode_service_request(
             data->cancellationRequest = false;
             len += decode_tag_number_and_value(
                 &apdu[len], &tag_number, &len_value);
-            data->issueConfirmedNotifications = len_value;
+            data->issueConfirmedNotifications = decode_boolean(len_value);
         } else {
             data->cancellationRequest = true;
         }
--- a/src/bacnet/cov.c
+++ b/src/bacnet/cov.c
@@ -489,9 +489,7 @@ int cov_subscribe_decode_service_request(
                 data->cancellationRequest = false;
                 len += decode_tag_number_and_value(
                     &apdu[len], &tag_number, &len_value);
-                data->issueConfirmedNotifications =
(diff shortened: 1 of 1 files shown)


skarg:
Thank you for finding the problem!

Please change the solution to use `bool decode_boolean(uint32_t len_value)` instead of len_value directly.

Also, please update `cov_subscribe_property_decode_service_request()` function. It has the same problem.

src/bacnet/cov.c
```diff
@@ -654,9 +652,7 @@ int cov_subscribe_property_decode_service_request(
             data->cancellationRequest = false;
             len += decode_tag_number_and_value(
                 &apdu[len], &tag_number, &len_value);
-            data->issueConfirmedNotifications =
-                decode_context_boolean(&apdu[len]);
-            len++;
```
stevedh, line 659: I am not really sure why this is being incremented here.  Wouldn't len be advanced past this option by decode_tag_number_and_value?

skarg: I agree, len shouldn't be incremented again. Seems like a copy/pasta error.

skarg (approved):

skarg:
Digging in a little deeper, I think the device emitting the COVSubscribe is wrong.  

From the BACnet standard:
```
20. ENCODING BACnet PROTOCOL DATA UNITS
...
20,2.3 Encoding of a Boolean Value
Context-tagged Boolean primitive data shall contain one contents octet. The value of this octet shall be B'00000000' if the value
to be encoded is FALSE or B'00000001' if the value to be encoded is TRUE.

Example: Context-tagged Boolean value
ASN.1 = [2] BOOLEAN
Value = TRUE
Context Tag = 2
Encoded Tag = X'29'
Encoded Data = X'01'
NOTE: The Boolean datatype differs from the other datatypes in that the encoding of a context-tagged Boolean value is not the
same as the encoding of an application-tagged Boolean value. This is done so that the application-tagged value may be encoded
in a single octet, without a contents octet. While this same encoding could have been used for the context-tagged case, doing
so would require that the context be known in order to distinguish between a length or a value in the length/value/type field.
This was considered to be undesirable.
```

stevedh:
I think you're right, the generating device is generating a broken encoding.  I was thrown off because WireShark at least seems to decode it right. Love those special cases in there ... 

Thanks for your help!

PR #446, opened 2023-07-10
Fix [bug#79] Out of bounds jump in h_apdu.c:apdu_handler

skarg:


src/bacnet/basic/service/h_apdu.c
anthony-crystalpeak, line 1: Same code pattern here

PR #449, opened 2023-07-12
Add trendlog test and zephyr specific trendlog

Michail-Antropov:


skarg:
BACnet Zephyr RTOS ports has moved to https://github.com/bacnet-stack/bacnet-stack-zephyr

PR #491, opened 2023-09-11
Fix wrong calculation of frames in bacapp_data_len

bakmaria:
In cases when 2 sets of opening & closing tags (first opening tag context tag is 3, second is 0) were sent, one of the frames wasn't added, and the resulted sum was wrong

### DIFF
--- a/test/bacnet/bacapp/src/main.c
+++ b/test/bacnet/bacapp/src/main.c
@@ -819,8 +819,10 @@ static void testBACnetApplicationDataLength(void)
     apdu_len += len;
     len = encode_opening_tag(&apdu[apdu_len], 0);
     apdu_len += len;
+    test_len += len;
     len = encode_closing_tag(&apdu[apdu_len], 0);
     apdu_len += len;
+    test_len += len;
     len = encode_closing_tag(&apdu[apdu_len], 3);
     apdu_len += len;
     /* verify the length of the data inside the opening/closing tags */


skarg (approved):

src/bacnet/bacapp.c
```diff
@@ -1544,8 +1544,9 @@ int bacapp_data_len(
                 opening_tag_number_counter = 1;
                 total_len_enable = false;
             } else if (tag_number == opening_tag_number) {
-                total_len_enable = true;
```
skarg, line 1547: `total_len_enable = true;` should not be deleted

skarg (changes requested):

skarg (approved):

test/bacnet/bacapp/src/main.c
```diff
@@ -811,6 +811,21 @@ static void testBACnetApplicationDataLength(void)
...
     len = bacapp_data_len(&apdu[0], apdu_len, PROP_REQUESTED_SHED_LEVEL);
     zassert_equal(test_len, len, NULL);
+
+    /* 7. context opening & closing tag */
+    test_len = 0;
+    apdu_len = 0;
+    len = encode_opening_tag(&apdu[apdu_len], 3);
+    apdu_len += len;
+    len = encode_opening_tag(&apdu[apdu_len], 0);
+    apdu_len += len;
```
skarg, line 821: please add `test_len += len;`

test/bacnet/bacapp/src/main.c
```diff
@@ -811,6 +811,21 @@ static void testBACnetApplicationDataLength(void)
...
+    len = encode_closing_tag(&apdu[apdu_len], 0);
+    apdu_len += len;
```
skarg, line 823: please add `test_len += len;`

skarg (changes requested):

PR #494, opened 2023-09-12
Remove depricated functions from ports/zephyr

Michail-Antropov:


ports/zephyr/event.c
```diff
@@ -1113,13 +1142,14 @@ int event_notify_decode_service_request(
                     default:
                         return -1;
                 }
-                if (decode_is_closing_tag_number(
-                        &apdu[len], (uint8_t)data->eventType)) {
+                if (bacnet_is_closing_tag_number(&apdu[len], apdu_len - len,
+                        (uint8_t)data->eventType, NULL)) {
```
skarg, line 1146: bacnet_is_closing_tag_number() returns the actual length of the closing tag in the 4th argument, and should be used (len++ assumes it is tag len of 1).

skarg (changes requested):

PR #511, opened 2023-10-12
Ability to define 'BAC_NO_PASSWORD' to ignore password for DCC and RD

jd-imi:


src/bacnet/basic/object/device.c
```diff
@@ -551,10 +551,14 @@ bool Device_Reinitialize(BACNET_REINITIALIZE_DEVICE_DATA *rd_data)
         protection, the service request shall be denied if the parameter
         is absent or if the password is incorrect. For those devices that
         do not require a password, this parameter shall be ignored.*/
+#ifdef BAC_NO_PASSWORD
```
skarg, line 554: Thanks for the fix! However, I think it would be a better example if you:

- add setter for reinit ASCII password (i.e. C string): `void Device_Reinitialize_Password_Set(const char *password);`
- check that Reinit_Password is not NULL in Device_Reinitialize() function.

Same with DCC.
-

PR #568, opened 2024-02-05
Add missing object functions to analog inputs and values

DocSepp:
In this PR I added the following functions to `Analog_Input` and `Analog_Value` objects:
- Name_Set
- Units
- Units_Set
- Description
- Description_Set

For this I needed to add `Object_Name` and `Description` properties to the data structures used by both. The name returned by the `Object_Name` functions was hard-coded to just return "Analog Input" or "Analog Value" respectively with the object instance. Now this will only be returned if the object name has not been set manually by a server application.

When a client would execute a read property request for the `Description` property the behaviour was to just return the object name instead. Now it will return the string stored in the `Description` property of the data structure.

All the newly implemented functions are already declared in the respective headers, so I think they should actually be available to users of this library. I also implemented the Units and Units_Set functions (that are declared in the headers as well) but I am not sure if the units for analog inputs and values should be hard-coded to `UNITS_PERCENT` and `UNITS_NO_UNIT`. 

What do you think?

### DIFF
--- a/src/bacnet/basic/object/ai.c
+++ b/src/bacnet/basic/object/ai.c
@@ -280,6 +280,40 @@ unsigned Analog_Input_Event_State(uint32_t object_instance)
     return state;
 }
 
+/**
+ * @brief For a given object instance-number, returns the description
+ * @param  object_instance - object-instance number of the object
+ * @return description text or NULL if not found
+ */
+char *Analog_Input_Description(uint32_t object_instance)
+{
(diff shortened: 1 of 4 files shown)


DocSepp:
Forgot to change MAX_ANALOG_VALUES to MAX_ANALOG_INPUTS when copying the code between av.c and ai.c

skarg (approved):

PR #624, opened 2024-04-23
Devel/persistent storage device hook master

metelik:
The hook is to be called after the Object_Write_Property handler and its implementation is optional and left to the stack's user.
The purpose is to enable support for relative permanence requirement by ASHRAE 135-2020 i.e.:
12.21.8 Recipient_List
This property, of type BACnetLIST of BACnetDestination, shall convey a list of zero or more recipient destinations to which
notifications shall be sent when event-initiating objects using this class detect the occurrence of an event. These recipient
destinations are intended to be **relatively permanent,** do not expire, and shall be maintained through a power failure or device
"restart." The destinations themselves define a structure of parameters that is summarized in Table 12-25.

It also is quite handy to cache the client settings to objects supporting intrinsic reporting.


skarg:
The change - in essence - adds two WriteProperty calls, the second dependent on the first succeeding.  Would this not be the equivalent of a developer creating their own object WriteProperty function and calling the typical WriteProperty within it?
```
bool Analog_Value_Write_Property_Store(
    BACNET_WRITE_PROPERTY_DATA * wp_data)
{
    bool status = false;

    status = Analog_Value_Write_Property(wp_data);
    if (status) {
        /* store some data */
    }

    return status;
}
````
The variety of non-volatile parameters and values and log data and other property values - in products - is something every product developer has to determine. I'm not sure the mechanism proposed is a good way to do it, as the same result can be accomplished by simply modifying the device object table handler for each WriteProperty as I indicated.

Note: I've been working on a [Device object refactoring](https://github.com/bacnet-stack/bacnet-stack/tree/feature/refactor-device-object-handling) that moves most of the Device object handling into the bacnet/basic/services realm, and a developer only needs to configure a few things - like the device object table - to get the generic device functionality.

Another concern - and one experienced on MS/TP products in particular - is the amount of time it takes to save data to certain types of non-volatile memory - which can drop a device from the token ring during the Data-Expecting-Reply frame while the client is waiting for storage to complete prior to sending the SimpleAck from WriteProperty.  An RTOS that runs the MSTP state machine and UART can fix this - but sometimes the storage is the microcontroller internal flash and the microcontroller disables all other activity while saving.

metelik:
Good day Sr,

Ad >> Would this not be the equivalent of a developer creating their own object WriteProperty function and calling the typical WriteProperty within it too.
Yes it would be the equivalent. I in fact was considering custom WP handlers. Whereas it can be done exactly as you outlined the intentions were not that apparent. The more glad I am consulting the 'eventual change.

Ad >> The variety of non-volatile parameters and values and log data and other property values - in products - is something every product developer has to determine.
Ineed, hence the proposed API change would be purely optional. The storage backends would vary from product to product.

Well, there would definitely be pros and cons.  Main ideas behind the approach taken by us was a uniformity as well as having it coded the logic of _if Write property succeeded; then Try storing it._ in one place rather than have the same condition in a number of custom WP handlers.

I have quickly re-factored our code to accommodate your suggestion and it works as anticipated. I would not get offended if my proposal gets rejected and I am ready to re-factor ;) What your call would be Mr.?

P. S. 
We would have few more extensions to share i.e.: https://github.com/se-apc/bacnet-stack/pull/39 .
I also scribed the CHANGE_OF_STATE algorithm along with intrinsic reporting for BI and BV objects.

metelik:
I deem we can close this one as we reached the consensus that the change would not be worth pursuing.

PR #686, opened 2024-07-02
Devel/netport modernisation

metelik:
- Adding the IP_DHCP_Enable optional property
- Adding FD_BBMD_Address & FD_Subscription_Lifetime for BIP6 network type
- Adding IPv6_Auto_Addressing_Enable property
- FD_BBMD_Address & FD_Subscription_Lifetime writable for the 'Foreign mode' as per **12.56.37** _This property, of type BACnetHostNPort, indicates the BBMD with which the local device is to register as a foreign device
when BACnet_IP_Mode is FOREIGN**. This property shall be present and writable if BACnet_IP_Mode is FOREIGN**.
If this property is writable, then a successful write to this property shall set the Changes_Pending property to TRUE. A value
written to this property shall become effective when the device receives a ReinitializeDevice service request with a
'Reinitialized State of Device' of ACTIVATE_CHANGES or WARMSTART._ 
- Some revolution in conditional compilation - feedback appreciated
- Introduced BBMD_CLIENT in addition to BBMD_ENABLED for eventual configuration not as a router but as a foreign device

Note 1: I am not certain as to the use of the compilation flags. Changes are the result of my work w/ the code and are mainly empirical

Note 2: The writability of the fdb_bbmd_address in the current state is impaired please see https://github.com/bacnet-stack/bacnet-stack/discussions/670  and https://github.com/se-apc/bacnet-stack/commit/526a72b8fccd8bc9846e20a44022679511eaddb6 . Another PR will follow. As always all the feedback is well appreciated.


src/bacnet/basic/object/netport.c
```diff
@@ -36,6 +36,7 @@
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
+#include "bacnet/config.h"
```
skarg, line 39: bacnet/bacdef.h already includes bacnet/config.h file as a basis, so adding bacnet/config.h to the start of these library is no longer needed (but bacdef.h is needed).

skarg (changes requested):

skarg (approved):

PR #687, opened 2024-07-04
Network priority fix for test 10.1.2

GauiStori:
This patch fixes test 10.1.2. I don't like the code but I need to fix the BTS errors.
The first two bytes of pdu are cut off before being sent to apdu_handler where they are needed
to set the response priority.


src/bacnet/basic/npdu/h_npdu.c
```diff
@@ -43,6 +43,7 @@
 static uint16_t Local_Network_Number;
 static uint8_t Local_Network_Number_Status = NETWORK_NUMBER_LEARNED;
 
+extern uint8_t priority;  /* Fixing test 10.1.2 Network priority */
```
skarg, line 46: Please use the same pattern as Local_Network_Number (naming, static scope) and a 'getter' function so as to not use a global variable 'priority'.

src/bacnet/basic/service/h_apdu.c
```diff
@@ -439,6 +439,14 @@ uint16_t apdu_decode_confirmed_service_request(uint8_t *apdu, /* APDU data */
         service_data->max_segs = decode_max_segs(apdu[1]);
         service_data->max_resp = decode_max_apdu(apdu[1]);
         service_data->invoke_id = apdu[2];
+        if (apdu[3] == 0x0c) /* Read property */
```
skarg, line 442: please apply to ALL confirmed service requests, not just ReadProperty.

skarg (changes requested):

src/bacnet/basic/service/h_apdu.c
GauiStori: Thanks. I hope I have fixed this correctly now.
The atmega168 port is fixed.

GauiStori:
Thanks Steve. Sorry for the error.

src/bacnet/basic/npdu/h_npdu.c
```diff
@@ -252,6 +252,10 @@ void npdu_handler(BACNET_ADDRESS *src, uint8_t *pdu, uint16_t pdu_len)
                     /* ConfirmedBroadcastReceived */
                     /* then enter IDLE - ignore the PDU */
                 } else {
+                    if ((pdu[apdu_offset] & 0xF0) == PDU_TYPE_CONFIRMED_SERVICE_REQUEST)
```
skarg, line 255: The NPDU is already decoded, so let's use that data to store the network priority:
```
    if (npdu_data.data_expecting_reply) {
         apdu_network_priority_set(npdu_data.priority);
    } else {
         apdu_network_priority_set(MESSAGE_PRIORITY_NORMAL);
    }
```

ports/atmega168/apdu.c
```diff
@@ -41,6 +41,26 @@
 #include "bacnet/bacdcode.h"
 #include "bacnet/basic/services.h"
 
+static int Local_Network_Priority; /* Fixing test 10.1.2 Network priority */
```
skarg, line 44: We should store the smaller uint8_t datatype that we're using in get/set.

src/bacnet/basic/service/h_apdu.c
```diff
@@ -53,6 +53,7 @@
 static uint16_t Timeout_Milliseconds = 3000;
 /* Number of APDU Retries */
 static uint8_t Number_Of_Retries = 3;
+static int Local_Network_Priority; /* Fixing test 10.1.2 Network priority */
```
skarg, line 56: We should store the smaller uint8_t datatype that we use in the get/set.

skarg (changes requested):

src/bacnet/basic/npdu/h_npdu.c
GauiStori: Ahh. Thanks. Then my implementation must have been wrong since I didn't check the expect reply bit.

ports/atmega168/apdu.c
GauiStori: Fixed

src/bacnet/basic/service/h_apdu.c
GauiStori: Fixed

GauiStori:
Thanks Steve. The last two commits should have fixed the issues.

skarg (approved):
Thank you for persevering my multiple pass approach to code reviews!

PR #688, opened 2024-07-10
Bugfix/bacdecode and bacapp hostnport

metelik:
A bugfix colllateral to #686 . Please review. I am uncertain whether this is the correct fix. The deliberations are detailed in the question section: https://github.com/bacnet-stack/bacnet-stack/discussions/670 .
Are we sure that a tag cannot be both: context specifix and opening. I wasn't able to pluck it out from ASHRAE 135-2020....


skarg:
Technically, yes, opening and closing tags are context tags as there are no application tags that are opening or closing tags. However, in parsing, it is important to distinguish between opening/closing (if we find an opening, we need to find a closing), and the context tagged application data or application tagged data or no data that falls between the opening and closing tags.

The problem in the code is the bacapp decode parsing, where skipping over the opening and closing and looking for the context tagged application data - when it is sequenced - is not handled very well.

The BACnet standard indicates FD_BBMD_Address and BACnet_IP_Global_Address are of type BACnetHostNPort, defined in clause 21 as:
```
    BACnetHostNPort ::= SEQUENCE {
        host [0] BACnetHostAddress,
            BACnetHostAddress ::= CHOICE {
                none [0] NULL,
                ip-address [1] OCTET STRING,
                -- 4 octets for B/IP or 16 octets for B/IPv6
                name [2] CharacterString
                -- Internet host name (see RFC 1123)
            }
        port [1] Unsigned16
    }
```

The bacapp_context_tag_type() function fails to choose an appropriate context that can be decoded where BACNET_APPLICATION_TAG_HOST_N_PORT should probably be a BACnetHostAddress (BACNET_APPLICATION_TAG_BACNET_HOST_ADDRESS) for case 0. But what to do about the next chunk (port[1]) in the sequence?
```
        case PROP_FD_BBMD_ADDRESS:
        case PROP_BACNET_IP_GLOBAL_ADDRESS:
            switch (tag_number) {
                case 0:
                    tag = BACNET_APPLICATION_TAG_HOST_N_PORT;
                    break;
                default:
                    break;
            }
```

I started a new branch for this issue.  I'm currently adding unit testing for the bacapp.c module that should show the functions that are broken for the specific complex data. I've also with refactored BACnetHostAddress encode/decode into their own functions assuming that we'll be using them. When I get it 'passing' I'll reference this issue.

skarg:
The fix was simpler than I thought.  The unit test was using `bacapp_decode_generic_property()` where the RP and RPM code use `bacapp_decode_known_property()` which includes special cases to decode complex property values. I changed the unit test to use `bacapp_decode_known_property()` and added a special case and the pseudo-application tag lookup using `bacapp_known_property_tag()` for the network port properties, and viola, the unit test is now passing.

See pull request [#700] for details.  Does this work in your application?

PR #689, opened 2024-07-10
Devel/bi bv intrinsic reporting change of state

metelik:
[Adding implementation of CHANGE_OF_STATE algorithm as per 13.3.2 as per ASHRAE 135-2020.

Just the early heads-up. I manually merged it from ours. May require re-work.


src/bacnet/basic/object/bi.c
```diff
@@ -825,6 +858,106 @@ int Binary_Input_Read_Property(BACNET_READ_PROPERTY_DATA *rpdata)
...
+            apdu_len = encode_application_bitstring(&apdu[0], &bit_string);
+            break;
+
+        case PROP_NOTIFY_TYPE:
+            apdu_len = encode_application_enumerated(
+                &apdu[0], pObject->Notify_Type ? NOTIFY_EVENT : NOTIFY_ALARM);
+            break;
+
+        case PROP_EVENT_TIME_STAMPS:
+            /* Array element zero is the number of elements in the array */
```
skarg, line 915: Please use the BACnetARRAY encoding helper function bacnet_array_encode() for the Event_Time_Stamps property - see example in ai.c file.
```
            apdu_len = bacnet_array_encode(
                rpdata->object_instance, rpdata->array_index,
                Analog_Input_Event_Time_Stamps_Encode,
                MAX_BACNET_EVENT_TRANSITION, apdu, apdu_size);
            if (apdu_len == BACNET_STATUS_ABORT) {
                rpdata->error_code =
                    ERROR_CODE_ABORT_SEGMENTATION_NOT_SUPPORTED;
            } else if (apdu_len == BACNET_STATUS_ERROR) {
                rpdata->error_class = ERROR_CLASS_PROPERTY;
                rpdata->error_code = ERROR_CODE_INVALID_ARRAY_INDEX;
            }
```

CMakeLists.txt
```diff
@@ -533,6 +533,9 @@ target_compile_definitions(
   ${PROJECT_NAME}
   PUBLIC
   BACNET_PROTOCOL_REVISION=${BACNET_PROTOCOL_REVISION}
+  INTRINSIC_REPORTING=1
+  BINARY_INPUT_INTRINSIC_REPORTING=1
```
skarg, line 537: I like BINARY_INPUT_INTRINSIC_REPORTING and BINARY_VALUE_INTRINSIC_REPORTING for more granular approach. However, INTRINSIC_REPORTING already exists and implies that all the objects intrinsic reporting features are enabled.

Suggestion: use INTRINSIC_REPORTING in Makefile or CMakeLists.txt, and in the bv.h and bi.h file we have another ifdef:
```
#if defined(INTRINSIC_REPORTING)
    #define BINARY_INPUT_INTRINSIC_REPORTING
#endif
```

src/bacnet/basic/object/bi.c
```diff
@@ -34,11 +39,26 @@ struct object_data {
     bool Present_Value : 1;
     bool Polarity : 1;
     bool Write_Enabled : 1;
+    unsigned Event_State:3;
     uint8_t Reliability;
     const char *Object_Name;
     const char *Active_Text;
     const char *Inactive_Text;
-    const char *Description;
+const char *Description;
```
skarg, line 47: Maybe run clang-format over the file to make the format consistent?  For tables where clang doesn't play nice (i.e. property lists) you can wrap in `/* clang-format off */` and `/* clang-format on */`

src/bacnet/basic/object/bv.c
```diff
@@ -830,6 +867,103 @@ int Binary_Value_Read_Property(BACNET_READ_PROPERTY_DATA *rpdata)
...
+            apdu_len = encode_application_bitstring(&apdu[0], &bit_string);
+            break;
+
+        case PROP_NOTIFY_TYPE:
+            apdu_len = encode_application_enumerated(
+                &apdu[0], pObject->Notify_Type ? NOTIFY_EVENT : NOTIFY_ALARM);
+            break;
+
+        case PROP_EVENT_TIME_STAMPS:
+            /* Array element zero is the number of elements in the array */
```
skarg, line 921: Please use the BACnetARRAY encoding helper function bacnet_array_encode() for the Event_Time_Stamps property - see example in ai.c file.

skarg (changes requested):

CMakeLists.txt
metelik: Excellent suggestion. How about #if defined(INTRINSIC_REPORTING) && (BINARY_VALUE_INTRINSIC_REPORTING) approach?

src/bacnet/basic/object/bi.c
metelik: Apologies for that... Forgot to %=G in my vim ;)

src/bacnet/basic/object/bv.c
metelik: Will do.

CMakeLists.txt
skarg: Since INTRINSIC_REPORTING has the expectation of enabling NC + AI + AV event notifications, it would break builds - for developers using INTRINSIC_REPORTING solely.  

Maybe change the object specific defines to require a 1, or a 0, using 1 with not defined, and add in each .h file. Even better would be to couple it with INTRINSIC_REPORTING, and then only use the object-specific define inline in code.
```
#ifdef INTRINSIC_REPORTING
#ifndef  BINARY_INPUT_INTRINSIC_REPORTING
#define BINARY_INPUT_INTRINSIC_REPORTING 1
#endif
#endif
```

metelik: O re-worked it so we would have a working code at which we could look. It's backwards compatible with the sole use of INTRINSIC_REPORTING flag. Next step I would remove the flags from CMakeLists.txt. I still am unsure  where would be the best place to put the BINARY_[INPUT|VALUE]_INTRINSIC_REPORTING=0  for a highlight to the users with perhaps a relevant comment.

src/bacnet/basic/object/bi.c
metelik: Re-worked.

src/bacnet/basic/object/bi.c
metelik: Re-worked.

src/bacnet/basic/object/bv.c
metelik: Re-worked.

skarg (approved):

PR #690, opened 2024-07-10
Add response to unknown network layer message types. 

GauiStori:
This patch should solve test 10.2.2.7.2. As far as I know it sends the correct reject message. Please take a look if interested. I will continue verifying it.


src/bacnet/basic/npdu/h_npdu.c
```diff
@@ -194,6 +195,15 @@ static void network_control_handler(BACNET_ADDRESS *src,
         default:
             break;
     }
+/* FIXME GG The following doesn't compile for the microcontroller ports. An appropriate ifdef is needed.
+Using MSTP so far */
+#ifndef BACDL_MSTP
+    if (npdu_data->network_message_type >= 0x14 && npdu_data->network_message_type < 0x80)
```
skarg, line 201: The magic numbers 0x14 and 0x80 are easier to maintain if they exist in the bacenum.h file in the BACNET_NETWORK_MESSAGE_TYPE enumeration:
```
    NETWORK_MESSAGE_ASHRAE_RESERVED_MIN = 20,
    NETWORK_MESSAGE_ASHRAE_RESERVED_MAX = 127,
```

skarg (changes requested):
I concur that the BACDL_MSTP define is not an appropriate ifdef. Seems more BTL tests are added that don't track with protocol-revision values, so it probably makes sense to fix in all the ports

src/bacnet/basic/npdu/h_npdu.c
GauiStori: Fixed, thanks

GauiStori:
This is now fixed for all ports.

src/bacnet/basic/npdu/h_npdu.c
```diff
@@ -194,6 +195,11 @@ static void network_control_handler(BACNET_ADDRESS *src,
         default:
             break;
     }
+    if (npdu_data->network_message_type >= NETWORK_MESSAGE_ASHRAE_RESERVED_MIN && npdu_data->network_message_type <= NETWORK_MESSAGE_ASHRAE_RESERVED_MAX)
```
skarg, line 198: please run clang-format on this file for the changes (80 column limit will be used)
please add extra parenthesis so developer don't have to think about precedence.

skarg (changes requested):

skarg:
Rather than adding the s_router.c module to any examples that need it, I would prefer that you create a new public API in h_npdu called npdu_send_reject_message_to_network() with appropriate arguments, similar to npdu_send_network_number_is().

GauiStori:
Thanks. This looks better.

ports/xplained/Makefile
```diff
@@ -59,6 +59,7 @@ BASICSRC = $(BACNET_BASIC)/tsm/tsm.c \
 	$(BACNET_BASIC)/sys/ringbuf.c \
 	$(BACNET_BASIC)/sys/mstimer.c \
 	$(BACNET_BASIC)/npdu/h_npdu.c \
+	$(BACNET_BASIC)/npdu/s_router.c \
```
skarg, line 62: please undo the changes to the ports as they are no longer needed

ports/at91sam7s/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ set(BACNET_PROJECT_SOURCE
     ${LIBRARY_BACNET_BASIC}/service/h_dcc.c
     ${LIBRARY_BACNET_BASIC}/service/h_apdu.c
     ${LIBRARY_BACNET_BASIC}/npdu/h_npdu.c
+	${LIBRARY_BACNET_BASIC}/npdu/s_router.c
```
skarg, line 125: it will save me time in the future if these lines can be aligned

skarg (changes requested):

GauiStori:
That I should have figured out :)

ports/at91sam7s/CMakeLists.txt
GauiStori: Removed. But I promise to remember.

GauiStori:
Thanks for the reviews Steve. Is there any comment that I have missed responding to in this PR? Not pushing. Please take your time.

skarg (approved):

PR #691, opened 2024-07-10
Set time

GauiStori:
This is an implementation for timesync that works on systems supporting
time(), localtime() and localtime_r() functions.

It uses an offset parameter from the system clock.

I assume I shall add ifdefs to the code for it to be merged.


skarg (changes requested):
datetime-init() should only reside in a specific 'port' - custom for whichever platform is being used or custom for the project you are developing - and not in the general application which may or may not be a POSIX port.

Please seconds or millseconds or minutes for storing a 'difference amount' and follow the naming pattern 'handler_timesync_offset_set()' and 'handler_timesync_offset()' for the set/get functions.  

The handler would be more maintainable if a callback was added to set the date/time in the custom port, and allow the developer to use the get/set in their own function.

GauiStori:
Thanks Steve. I changed the offset to int32, it may be positive or negative. Forgot that float is expensive on microcontrollers. The modulus and division is in fact also expensive in the datetime_local function. Would a struct timeval be cheaper?
I haven't tested the code and I need to think a bit on how to implement the callback will change to a PR when done.

GauiStori:
Now all requests are fixed.
I got a hint from Edward that the UTC_offset should be the difference between "Solar time" and UTC, not Local time and UTC. The standard says "Local standard time", I assume that means "Solar time".

I have tried to test the mingw port on my Linux with Wine but it doesn't work as expected. It responds to WhoIs and Timesync requests but not nothing else.
My VS2019 doesn't respond. It compiles and runs but doesn't show up in WhoIs.

I don't have any MacOS. 

Hope this is ready for review despite these problems. I will try to fix these ASAP.

skarg:
Suggestions:
1. move the get/set of offset into h_ts.c module as common code
2. add callback and get/set of callback function into h_ts.c module as common code
3. don't add a callback set to the example applications that receive the time-sync, so no need to create a set-time for linux/macos/bsd/windows.

Callback in h_ts.h could look like:
typedef void (*handler_timesync_set_callback)(
    BACNET_DATE *bdate,
    BACNET_TIME *btime,
    bool utc);

Search the stack library for examples of other callbacks, and add a "set" function to set the callback (in h_ts.c module).

GauiStori:
Thanks Steve. I hope this is near to what you mean. I have tested the linux port and it works nicely but I don't know why the mstp port fails. Tried to find a solution but failed.

src/bacnet/basic/service/h_ts.c
```diff
@@ -276,3 +291,13 @@ void handler_timesync_init(void)
     }
 }
 #endif
+
+void handler_timesync_set_callback_set(
+    handler_timesync_set_callback_t cb)
+{
+    if (cb) {
```
skarg, line 298: There is no need to check the cb for NULL if we are going to set the callback to NULL when the cb is NULL.

ports/linux/datetime-init.c
```diff
@@ -15,9 +15,54 @@
...
+}
+
+/**
+ * @brief Set offset from the system clock.
+ * @param bdate BACnet Date structure to hold local time
+ * @param btime BACnet Time structure to hold local time
+ * @param utc - True for UTC sync, False for Local time
+ * @return True if time is set
+ */
+void datetime_timesync(
```
skarg, line 35: please add this public API to src/bacnet/datetime.h file at the bottom with the other "implementation agnostic functions - create your own!"

src/bacnet/basic/service/h_ts.c
```diff
@@ -66,6 +66,20 @@ static void show_bacnet_date_time(BACNET_DATE *bdate, BACNET_TIME *btime)
 }
 #endif
 
+static handler_timesync_set_callback_t handler_timesync_set_callback;
+
+static int32_t Time_Offset; /* Time offset in ms */
+
+int32_t handler_timesync_offset()
```
skarg, line 73: Please add 'void' to function parameters. 
Please add JavaDoc style doxygen function header comments to new public API functions.

skarg (changes requested):

src/bacnet/basic/service/h_ts.c
GauiStori: Thanks Steve. Fixed in 898923b9e0f690f3dc63f5ea0c1ee7d8316a041e

src/bacnet/basic/service/h_ts.c
GauiStori: Thanks. Fixed in 898923b9e0f690f3dc63f5ea0c1ee7d8316a041e

ports/linux/datetime-init.c
```diff
@@ -85,5 +134,5 @@ bool datetime_local(BACNET_DATE *bdate,
  */
 void datetime_init(void)
 {
-    /* nothing to do */
+    handler_timesync_set_callback_set(&datetime_timesync);
```
skarg, line 137: Although the set of Linux projects that want their clock updated on a BACnet timesync is non-zero, there are probably a non-zero set of them that don't want that feature. 

Let's move this callback set to the server example main.c module, and add an empty dummy stub function of datetime_timesync() to the MacOS and Win32 ports datetime-init.c module.

skarg (changes requested):

skarg (approved):

ports/bsd/datetime-init.c
```diff
@@ -86,5 +101,6 @@ bool datetime_local(BACNET_DATE *bdate,
  */
 void datetime_init(void)
 {
+    handler_timesync_offset_set(0);
```
skarg, line 104: should the offset be configured per project instead of per port?

skarg (changes requested):

ports/bsd/datetime-init.c
GauiStori: Thanks. You are right.

skarg (approved):

PR #694, opened 2024-07-15
Bugfix/mstp not working windows

KishorDanfoss:
Fix for MSTP not working on windows.

- Calls master state machine repeatedly in case of state transition.
-  Change ReadTotalTimeoutConstant from 1 second to 1 ms. Now ReadFile() returns after 1 ms if there is no data to read instead of 1 second.
- Treply_delay is set to Default value.


src/bacnet/datalink/dlenv.c
```diff
@@ -643,7 +643,7 @@ void dlenv_init(void)
     if (pEnv) {
         dlmstp_set_max_master(strtol(pEnv, NULL, 0));
     } else {
-        dlmstp_set_max_master(127);
+        dlmstp_set_max_master(7);
```
skarg, line 646: The changes to DLENV appear to be for testing. Please use OS environment variables for changes to the values in dlenv.c module, or some other method (not hard coding).  The values in the 'else' are used when a value has not been specified.

src/bacnet/datalink/dlenv.c
```diff
@@ -655,7 +655,7 @@ void dlenv_init(void)
     if (pEnv) {
         dlmstp_set_mac_address(strtol(pEnv, NULL, 0));
     } else {
-        dlmstp_set_mac_address(127);
+        dlmstp_set_mac_address(5);
```
skarg, line 658: please revert this change - presumed it was for testing?

src/bacnet/datalink/dlenv.c
```diff
@@ -671,7 +671,7 @@ void dlenv_init(void)
         apdu_retries_set((uint8_t)strtol(pEnv, NULL, 0));
     }
     /* === Initialize the Datalink Here === */
-    if (!datalink_init(getenv("BACNET_IFACE"))) {
+    if (!datalink_init("COM14")) {
```
skarg, line 674: Please revert this change - presume it was for testing?

src/bacnet/datalink/mstp.c
```diff
@@ -1671,7 +1671,8 @@ void MSTP_Init(struct mstp_port_struct_t *mstp_port)
         if ((mstp_port->Tframe_abort < 6) || (mstp_port->Tframe_abort > 100)) {
             mstp_port->Tframe_abort = DEFAULT_Tframe_abort;
         }
-        if (mstp_port->Treply_delay > 250) {
+        if ((!mstp_port->Treply_delay) || mstp_port->Treply_delay > 250)
```
skarg, line 1674: I'd prefer to use the pattern of comparing the Treply_delay to a value (zero) instead of using the "!" operator.

Also, please fix the indents (there is a .clang-format in the root of the project that is used for the clang-format tool).

skarg (changes requested):

skarg (approved):

PR #696, opened 2024-07-16
Add check on apdu_len. Timeout if too long.

GauiStori:
BTS has a test on apdu length. The current code responds to telegrams with apdu_len > MAX_APDU.
This patch changes this to timeout.

### DIFF
--- a/src/bacnet/basic/service/h_apdu.c
+++ b/src/bacnet/basic/service/h_apdu.c
@@ -468,7 +468,9 @@ uint16_t apdu_decode_confirmed_service_request(uint8_t *apdu, /* APDU data */
                 return 0;
             }
         }
-        if (apdu_len == (len + 1)) {
+        if (apdu_len > MAX_APDU){
+            return 0;
+        } else if (apdu_len == (len + 1)) {
             /* no request data as seen with Inneasoft BACnet Explorer */
             *service_choice = apdu[len++];
             *service_request = NULL;


src/bacnet/basic/service/h_apdu.c
```diff
@@ -468,7 +468,9 @@ uint16_t apdu_decode_confirmed_service_request(uint8_t *apdu, /* APDU data */
                 return 0;
             }
         }
-        if (apdu_len == (len + 1)) {
+        if (apdu_len > MAX_APDU){
```
skarg, line 471: Please use 'len' instead of MAX_APDU (it should be the same (less the NPDU) when it gets into this function).

GauiStori: Thanks Steve. This time  I have to admit that I don't understand how to use len check if apdu_len is larger than MAX_APDU, 1476 bytes in my case.
Your existing code takes care of 
apdu_len >= len+1 but fails the test.

skarg: npdu_handler(pdu, pdu_len) -> where pdu_len <= MAX_MPDU, and MAX_MPDU = datalink header + MAX_PDU, and MAX_PDU = MAX_NPDU+MAX_APDU.

Hmm, I see the point. The datalink header and NPDU are not fixed sizes.

skarg (approved):

skarg:
GCC MS/TP build should now be fixed in the master branch - if you could merge with your branch?

PR #704, opened 2024-07-25
mstpcap for macOS

empeeoh:
Altered mstpcap bsd port to also work on macOS.

Mainly involved augmenting makefile and ports/bsd/rs485.c to account for macOS differences.

If unacceptable, could make a separate ports subdir just for Darwin/macOS and resubmit branch.

Thanks for considering!


skarg:
Thank you for the contribution. A couple things:

- we probably don't want dlmstp_linux.c in the ports/bsd folder.  It's only used by the apps/router and if we want to port it to BSD we can name it appropriately and change the Makefile to include the BSD specific file.
- mstpcap Makefile only uses the C files specified in the apps/mstpcap/Makefile.  Please don't add BACNET_LIB_TARGET.

empeeoh:
Thank you for the quick review Steve!

I’ve tried to implement your recommendations. I’ll re-sumbit the pull request with an amended final commit on that branch.

-I removed most of the files that were clumsily brought to the ports/bsd dir in an earlier commit since they are not needed. Notably dlmstp_linux.c dlmstp.c are gone and dlmstp_linux.h has been renamed dlmstp_bsd.h its declaration and definition of the shared_mstp_data struct is necessary for rs485.c to function.

-I removed the clumsiness I introduced to the makefile re: BACNET_LIB_TARGET per your guidance. I think i did that when I was still figuring out how the apps fit together w/ the larger lib.

Thanks again. I hope this is all appropriate etiquette.

Best,
	Mike O'Neill

PR #705, opened 2024-07-26
Mstpcap macos

empeeoh:
Altered mstpcap bsd port to also work on macOS.

Mainly involved augmenting makefile and ports/bsd/rs485.c to account for macOS differences.

If unacceptable, could make a separate ports subdir just for Darwin/macOS and resubmit branch.

Thanks for considering!


ports/bsd/rs485.c
```diff
@@ -0,0 +1,899 @@
...
+    // Failure path
+error:
+    if (fileDescriptor != -1) {
+        close(fileDescriptor);
+    }
+    
+    return -1;
+}
+
+// Given the file descriptor for a serial device, close that device.
```
skarg, line 831: Although it probably won't matter in this port, the rest of the library uses C comment blocks, not C++, as the library code needs to work with C89 compilers that don't support `//`.

skarg (changes requested):

ports/bsd/dlmstp_bsd.h
```diff
@@ -0,0 +1,178 @@
...
+#define DLMSTP_BSD_H
+
+#include "bacnet/datalink/mstp.h"
+#include <sys/types.h>
+#include <semaphore.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stddef.h>
+#include "bacnet/bacnet_stack_exports.h"
```
skarg, line 21: bacdef.h now includes bacnet/bacnet_stack_exports.h, so it can be removed

ports/bsd/rs485.c
```diff
@@ -0,0 +1,881 @@
...
+ * operation
+ * @author Steve Karg <skarg@users.sourceforge.net>
+ * @author Nikola Jelic <nikola.jelic@euroicc.com>
+ * @author Michael O'Neill <em.pee.oh@gmail.com>
+ * @date 2007, 2011, 2024
+ *
+ * @copyright SPDX-License-Identifier: MIT
+ */
+
+/* The module handles sending data out the RS-485 port */
```
skarg, line 13: no need to have these duplicitous comments, please remove

apps/mstpcap/Makefile
```diff
@@ -28,11 +28,9 @@ SRCS = main.c \
 	${BACNET_SRC_DIR}/bacnet/datalink/mstptext.c \
 	${BACNET_SRC_DIR}/bacnet/datalink/crc.c
 
-# This demo seems to be a little unique
+# This demo seems to be a little unique -skarg
```
skarg, line 31: no need to add skarg.

skarg (approved):

ports/bsd/rs485.c
empeeoh: duplicative or duplicitous? If you mean the latter, yes Line 13 is definitely redundant. But if you meant the former, I meant no offense and I can remove lines 7 and 6 and edit the dates on 8.

skarg: redundant - no offense taken. Less lines means less code to maintain.

empeeoh:
So, last neophyte Q about github pull-req etiquette. You approved the pull req. (Thank you!) but also had some further requested edits. Does that mean I should submit an entirely _new_ pull req.? Before, I'd done a re-review submittal, but that is now closed.

Also, going forward, would you prefer on updating the header to use a) SPDX and b) doxygen/Javadoc, that should be just for entirely new files, or a housekeeping upgrade for any file that is otherwise being edited?

skarg:
I'm a github neophyte as well, but most things were pretty straightforward.  
I approved since the 'comment' fixes are nice to have but not required. I didn't realize it would eliminate a workflow.  From my work in pull requests, I just edit the files in my branch and push again. You should not have to create a new pull request.

Since SPDX/doxygen are 'comments', I tend to ask only for new files. I have been spending time fixing the existing files.  However, updating existing files would be greatly appreciated.

PR #708, opened 2024-07-30
Disallow setting network number to 65535

ericwmiller:
ASHRAE 135-2020 says:

- 12.56.11 Network_Number: "The range for this property shall be 0 .. 65534."

- 6.1 Network Layer Service Specification: "A network number of X'FFFF' indicates that the message is to be broadcast "globally" to all devices on all currently reachable networks."


skarg:
For the command line applications, we usually allow command line arguments and values that are in a valid encoding range, even if they are 'invalid' according to the specification.  This enables command line applications to perform some negative testing within scripts. Is this a change that you need for your usage?

ericwmiller:
No, I can limit my usage to the range allowed by the standard.

When I was using the command line tools, I noticed that some values were
being rejected as out-of-range and others were not, so my thought was that
if the tool is going to reject invalid values, it should match the standard.

It's a tough line to draw for test tools, so I will defer to you as the
maintainer. Is there anything I need to do to remove my pull request?

skarg:
Yes, It is a tough line to draw for command line applications that double as test tools. Thanks for understanding. 

I recently opened up some of the ranges (i.e. target device = 4194303) in the command line applications to enable more use as testing tools for objects (i.e. network port, device) that use that value as a wildcard.

Issue #644, opened 2024-05-17
how to work bacnetip and mstp together 


cyberumut:
how do bacnetip and mstp work together? is this possible?  

i want to use the same table

skarg:
MS/TP and BACnet/IP can work together:

- BACnet/IP to MS/TP router: passes appropriate APDU with matching DNET or broadcasts to the downstream datalink
- BACnet/IP and MS/TP as separate non routed datalinks: originates messages to BACnet/IP or MS/TP without routing messages between them.

Issue #205, opened 2021-11-11
CMake build does not support BACDL_ALL


AlexanderWells-diamond:
There seems to be no way to tell CMake configuration to set BACDL_ALL macro, only a series of individual BACDL_* macros e.g. BACDL_ETHERNET and BACDL_BIP. 

I'm trying to use this library to support both Ethernet and BIP communication at runtime (I'm actually upgrading a server from the old bacnet4linux library). This means I must compile a library with both the Ethernet and BIP macros. However, enabling both of these through CMake configuration results in a failing build:

```
[  0%] Building C object CMakeFiles/bacnet-stack.dir/src/bacnet/basic/npdu/h_routed_npdu.c.o
/usr/bin/cc -DBACDL_BIP -DBACDL_ETHERNET -DBACNET_PROPERTY_LISTS -DBACNET_PROTOCOL_REVISION=19 
-DBACNET_STACK_STATIC_DEFINE -DBAC_ROUTING -DPRINT_ENABLED=1 -I/<cut>/bacnet-src/src 
-I/<cut>/bacnet-src/ports/linux  -MD -MT CMakeFiles/bacnet-stack.dir/src/bacnet/basic/npdu/h_routed_npdu.c.o 
-MF CMakeFiles/bacnet-stack.dir/src/bacnet/basic/npdu/h_routed_npdu.c.o.d 
-o CMakeFiles/bacnet-stack.dir/src/bacnet/basic/npdu/h_routed_npdu.c.o 
-c /<cut>/bacnet-src/src/bacnet/basic/npdu/h_routed_npdu.c
/<cut>/bacnet-src/src/bacnet/basic/npdu/h_routed_npdu.c: In function ‘routed_apdu_handler’:
/<cut>/bacnet-src/src/bacnet/basic/npdu/h_routed_npdu.c:208:41: error: ‘BVLC_ORIGINAL_UNICAST_NPDU’ undeclared 
(first use in this function)
         if (bvlc_get_function_code() != BVLC_ORIGINAL_UNICAST_NPDU) {
                                         ^
```

[...]

PR #717, opened 2024-08-13
Make it possible to select multiple datalinks

teksturi:
Before we could define all, one or none datalinks. Improve this so that we can actually just select multiple datalinks. Also now we can take out configure option BACDL_NONE. We just define that, if none is selected.

I choose to do most of the magic in config.h file so that people will not be depended from CMake. Note that I have not test zephyr part. I would be very happy if someone can test that as I do not understand that part yet.

Note that currently when selecting BACNET_STACK_BUILD_APPS and BACDL_BIP6 router-ipv6 will fail to build. This has already been there and will be fixed in other PR.

Releted issues:
  - https://github.com/bacnet-stack/bacnet-stack/issues/644
  - https://github.com/bacnet-stack/bacnet-stack/issues/205

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,33 +46,37 @@ option(
 
 option(
   BACDL_ETHERNET
-  "compile with ethernet support"
+  "compile with ethernet datalink support"
   OFF)
 
 option(
   BACDL_MSTP
-  "compile with mstp support"
+  "compile with mstp datalink support"
   OFF)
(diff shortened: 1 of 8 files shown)


skarg (approved):

nazaryev-cool:
Is there anything stopping you from merging these changes?

skarg:
Sorry, delays from traveling...

PR #718, opened 2024-08-13
Add and fix compiler warnings

teksturi:
Add some new compiler warnings which are used with some users. Currently we basically ignore those but this is lot easier for users of this project.

I also made it so that warnings should not get passed anymore in CI with CMake build. Our main matrix build also builds tests. I have to disable windows build for tests as it did not go through. It got like zillion errors.

### DIFF
--- a/ports/linux/bip6.c
+++ b/ports/linux/bip6.c
@@ -284,7 +284,7 @@ uint16_t bip6_receive(
     int max = 0;
     struct timeval select_timeout;
     struct sockaddr_in6 sin = { 0 };
-    BACNET_IP6_ADDRESS addr = { { 0 } };
+    BACNET_IP6_ADDRESS addr = { 0 };
     socklen_t sin_len = sizeof(sin);
     int received_bytes = 0;
     int offset = 0;
--- a/src/bacnet/getevent.c
+++ b/src/bacnet/getevent.c
@@ -65,6 +65,7 @@ size_t getevent_service_request_encode(
  * @param lastReceivedObjectIdentifier  Object identifier
  *
  * @return Bytes encoded.
+ * @deprecated Use getevent_apdu_encode() instead
  */
 int getevent_encode_apdu(uint8_t *apdu,
     uint8_t invoke_id,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,12 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macOS-latest]
-
+        project: [root, test]
+        exclude:
+          # Currently does not build. Need to be fixed at some point.
+          - os: windows-latest
+            project: test
+  
     steps:
     - uses: actions/checkout@v4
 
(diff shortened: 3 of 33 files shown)


src/bacnet/getevent.c
```diff
@@ -18,7 +18,7 @@
  * @param lastReceivedObjectIdentifier  Object identifier
  * @return Bytes encoded.
  */
-int getevent_apdu_encode(uint8_t *apdu,
+static int getevent_apdu_encode(uint8_t *apdu,
```
skarg, line 21: can getevent_apdu_encode() be added to header file instead?

src/bacnet/basic/object/structured_view.c
```diff
@@ -462,7 +462,8 @@ bool Structured_View_Represents_Set(
  * @param  object_instance - object-instance number of the object
  * @return number of Subordinate_List elements
  */
-unsigned int Structured_View_Subordinate_List_Count(uint32_t object_instance)
+static unsigned int Structured_View_Subordinate_List_Count(
```
skarg, line 465: Can these structured view functions (taking argument of object-instance) be added to header file instead?

ports/linux/bip6.c
```diff
@@ -284,7 +284,7 @@ uint16_t bip6_receive(
     int max = 0;
     struct timeval select_timeout;
     struct sockaddr_in6 sin = { 0 };
-    BACNET_IP6_ADDRESS addr = { { 0 } };
+    BACNET_IP6_ADDRESS addr = { { 0 }, 0 };
```
skarg, line 287: It is easier to maintain declaration initial value as `= { 0 };` pattern - is that possible in this case?

skarg (changes requested):

src/bacnet/getevent.c
teksturi: Of course. You know the best what should be moved to headers. Just say and I will move those ones.

ports/linux/bip6.c
teksturi: > It is easier to maintain declaration initial value as `= { 0 };` pattern - is that possible in this case?

Yeah that is better way in this. I do not know why I miss that. Thanks for review. If anyone else is wondering what this means: we are currently doing double braces like { { 0 } } compiler things we might actually want field initialize. Newer compilers does not however give warning if we do just { 0 } which is current idiom in C to do struct initializing.

I add couple more place where I found this same thing. Can be checked from commit after I push.

src/bacnet/getevent.c
teksturi: I notice that in header we already have getevent_encode_apdu() so now we would have

```c
    BACNET_STACK_EXPORT
    int getevent_apdu_encode(
        uint8_t *apdu,
        BACNET_OBJECT_ID *lastReceivedObjectIdentifier);

    BACNET_STACK_EXPORT
    int getevent_encode_apdu(
        uint8_t * apdu,
        uint8_t invoke_id,
        BACNET_OBJECT_ID * lastReceivedObjectIdentifier);
```

This looks strange as those are almost identical. But I do not feel yet that I can make call here. So please advice is this ok or not. I can change it anyway if needed.

src/bacnet/basic/object/structured_view.c
teksturi: Definetly.

src/bacnet/getevent.c
skarg: tl;dr; add DEPRECATED() to getevent_encode_apdu() declaration, include both declarations in header file

The original naming and features of the 'encode' of a service added the invoke ID instead of just initializing the APDU header separately.  I found when working in another project with this entanglement, that it is better to have separate service encode and service header encode.

The naming convention that is easiest for me to maintain is module_name+noun+verb, or even bacnet+module_name+noun+verb to help give a unique namespace.  The bacnet+module_name+noun+verb - for encoders and decoders would include apdu and apdu_size.  The shorter version, more for existing code, includes just apdu without size and can still be used, first with a call where apdu=NULL to get the size, and if size is okay, then call with apdu=buffer to fill the buffer.

teksturi: Can you please review that last commit do what you wanted.

skarg (approved):

PR #722, opened 2024-08-14
Force C89/C90 compatible and for test C99

teksturi:
Library seems to be made so that it is fully C89/C90 compatible. There where couple things which needed fixing but those where trivial. We should force compatible with CMake so no one breaks it by accident. If we at some point choose to move to C99 or even C11 then we can make it clear in CMake what version we support.

Tests where using code which is not totally C89/C90 compatible so C99 was chosen there. This is in my opinion totally acceptable as those are "just tests".


teksturi:
No is also good time to discuss do we want this library to be C89 compatible or is C99 fine also?

skarg (approved):

PR #729, opened 2024-08-16
Make clean build with MSVC /Wall

teksturi:
This PR makes project build cleanly with MSVC /Wall compiler option. 

There was mistake with previously made CI with warnings as error flag. It did not work but I found lot cleaner solution which works for MSVC and others. That is reason we also need to fix one -Wdeclaration-after-statement compiler warning.

Next by making CMake files for apps it lot easier for add example build parameters for different apps. Also new comers can start modifying CMakeLists.txt file and add new files to it pretty easily. That can also be starting point for someone else for their project. This is no means yet perfect and can be improved. At some point we can probably drop those makefiles and just use CMake for all things.

I then added /Wall and ignore warnings which we do not need or we need to fix later. Clean build is what matter most at this point. I also fixed easy to fix warning.


skarg:
I use CMake more and more for building on a variety of systems, so these additions are very welcome.

Note: We considered dropping Makefile builds previously. However, there are vendors using build environments that do not support CMake and requested that we continue the support of Makefile.

Note: The root Makefile is easy to maintain for pipeline integration and to have an easy method for repeating any failed pipeline steps.

skarg:
Could you split this pull request:
1. fixes/changes for the warnings
2. cmake in apps subdirectories (likely targeting next release v1.4.0?)

The current CMakeLists.txt doesn't account for nuance of routers (multiple datalinks ignoring BACDL) when building or tools like mstpcap that build regardless of BACDL, so those recipes will likely need some changes.

teksturi:
New PR made for warnings https://github.com/bacnet-stack/bacnet-stack/pull/740 cmake PR will come later.

PR #740, opened 2024-08-22
Make clean build with MSVC /Wall

teksturi:
There was mistake with previously made CI with warnings as error flag. It did not work but I found lot cleaner solution which works for MSVC and others. That is reason we also need to fix one -Wdeclaration-after-statement compiler warning.

I then added /Wall and ignore warnings which we do not need or we need to fix later. Clean build is what matter most at this point. I also fixed easy to fix warning.


CMakeLists.txt
```diff
@@ -138,6 +138,36 @@ if (CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "AppleCla
   add_compile_options(-Wno-float-conversion)
   add_compile_options(-Wno-missing-declarations)
   add_compile_options(-Wno-unused-but-set-variable)
+  elseif(MSVC)
```
teksturi, line 141: fix indentation

skarg (approved):

PR #744, opened 2024-08-23
Move some external files to external/ folder

teksturi:
It is lot easier to work with automatic formatters if we have external files in different folder. For some tools we can example just exclude external all together. This work is done so we can get example example clang-format to work easily in whole repo.

## Testing

Not tested as I do not have these dev environments.


ports/xplained/bacnet.cproj
```diff
@@ -240,32 +240,32 @@
           <ListValues>
             <Value>..</Value>
             <Value>../config</Value>
```
teksturi, line 242: todo

skarg (approved):

PR #745, opened 2024-08-23
Bugfix/freebsd port build

skarg:


ports/bsd/rs485.c
```diff
@@ -42,8 +42,14 @@
 
 #include "dlmstp_bsd.h"
 
-/*macOS-darwin includes*/
+#ifdef __APPLE__
 #include <IOKit/serial/ioss.h>
+#endif
+#if defined(__APPLE__) || defined(__darwin__)
+#include <IOKit/serial/ioss.h>
```
teksturi, line 49: This is now two times. Probably remove upper include

teksturi (approved):
Looks reasonable.

Issue #746, opened 2024-08-23
Add automatic formatter tools to repo


teksturi:
Plan is to to add new checks and formatters to repo so that code and files are checked easily locally and in CI.

Main thing is that we get clang-format to work cleanly in repo so we can start discard PRs what are not formatted. To make live easier plan is to use pre-commit software what runs clang-format. This way every commit can always be automatically formatted if dev wants that. Also everyone will always use same version of clang-format. Plan is also to start use .editorconfig file which does basic formatting for all files. So usually IDEs will example then know tabs/spaces, indentation level and trim trailing whitespaces.

This work needs couple PR so issue is created. It is easier if at least most of the external files are external/ folders. This way we can just block those. Also some manual fixes are needed before we run formatters to whole repo.

PR #753, opened 2024-08-28
Use pre-commit and editorconfig

teksturi:
To make life easier use pre-commit and editorconfig.

Editorconfig is common way to tell many different IDEs how files should be formatted. Things like line endings and that short of small things.

Pre-commit is Python based tool which many open-source projects already use. It has very easy interface to add new checks and formatters. Normally people will instruct people about these things but with pre-commit it is easy to run checks and formatters locally and in CI.

I do not yet want to run clang-format as that is lot bigger and other changes are not shown then. You could still test locally this by uncommenting clang from pre-commit config and run 

```shell
git add .pre-commit-config
pre-commit run --all-files
```

Relates to
  - https://github.com/bacnet-stack/bacnet-stack/issues/746

WIP: I test CI and really want this PR https://github.com/bacnet-stack/bacnet-stack/pull/744 to me merged before this


.clang-format
```diff
@@ -10,6 +10,7 @@ AllowShortLoopsOnASingleLine: false
 AlignEscapedNewlines: Left
 AlignArrayOfStructures: None
 PointerAlignment: Right
+InsertBraces: true
```
teksturi, line 13: @skarg is this ok to you?

skarg: Yes, InsertBraces is okay with me; saves me from running clang-tidy. 👍

skarg:
I've added `pre-commit/action@*` to the previously empty list of allowed actions.

skarg:
note: there is a `.git-blame-ignore-revs` file for tracking formatting changes (not complete or accurate for the entire history).

teksturi:
> note: there is a `.git-blame-ignore-revs` file for tracking formatting changes (not complete or accurate for the entire history).

Yeah I know. Problem is that in this repo everything is squashed together when merging. When squash is used it is way too dangerous imo to add that squashed commit to .git-blame-ignore-revs. I also do not actually think anyone is really using git history that way that is meaningless to add anything to ".git-blame-ignore-rev". This is because usually squashed commit make them way too hard to follow. Lot of things change with just couple lines of text. 

Also we can only add commits to  `.git-blame-ignore-revs` after we have merged them to master. Else commit shas will change.

teksturi:
This is ok to be merged on my part. If you @skarg need personal presentation about pre-commit or editorconfig please ask and we can make it happened at some point.

skarg (approved):

PR #757, opened 2024-08-29
Moved west manifest, zephyr folder, and ports/zephyr folders to anoth…

skarg:
…er repository bacnet-stack-zephyr


teksturi:
Wow. Big changes happening. Are you sure you want to do this? Always harder to maintain when something is in different repo. I'm of course totally ok with this but I'm also interested why this change. Could also be just totally own folder for these in this repo.

skarg:
Zephyr OS is rapidly moving and constantly introducing breaking changes, unlike Windows/Linux/BSD ports which move much more slowly and have, for the most part, had backward compatibility in mind.  Trying to keep up with the changes in this our repo with the changes in Zephyr OS and its own testing harness have been challenging at best, but more problematic for most developers that are contributing.  Moving the churn into [bacnet-stack-zephyr](https://github.com/bacnet-stack/bacnet-stack-zephyr) should place the churn there.

The latest change is that Zephyr OS 3.7.0 LTS is now released, but the west.yml in our library is for 3.6.0 and breaks on 3.7.0. Yes, we could just make a long lived branch that is bacnet-stack-1.3-zephyr-3.6 and bacnet-stack-1.3-zephyr-3.7 and try to maintain those, but we thought it would be easier to move the zephyr manifest from this repo and place it in another that is only focused on zephyr. I suppose time will tell if this is the right way to go.

teksturi:
Yeah it can always be moved back if needed.

gregshue (approved):

PR #758, opened 2024-08-30
Release/version 1.4.0 candidate

skarg:


teksturi:
I would still wait quite bit of this right? At least I would like to get couple things done before this. Const correctness would be nice that it is at least like 99% done. Needs couple PR for that.  Do we have any hurry on this one or did you just create like place holder for this?

skarg:
Thanks for letting me know pending plans for 'const' correctness, and I think there are some datalink changes pending?
I think this branch can wait some time (as much as you need). My plan is to be able to link the [bacnet-stack-zephyr](https://github.com/bacnet-stack/bacnet-stack-zephyr/) repo to a release.

PR #763, opened 2024-09-13
Devel/resolve todo buffer overrun master

metelik:
Resolving a potential buffer overrun - a ToDo leftover...


src/bacnet/basic/service/h_cov.c
```diff
@@ -271,34 +271,46 @@ static int cov_encode_subscription(
...
- *  @note This function needs some work to better handle buffer overruns.
  *  @param apdu [out] Buffer in which the APDU contents are built.
  *  @param max_apdu [in] Max length of the APDU buffer.
  *  @return How many bytes were encoded in the buffer, or -2 if the response
  *          would not fit within the buffer.
  */
+/* Maximume length for an encoded COV subscription  - 27 bytes for BACNET IP6
+ * 31 bytes for IPv4 (longest MAC) and lets round it up to the machine word
+ * alignment */
+#define MAX_COV_SUB_SIZE (32)
```
skarg, line 282: To avoid having to reserve buffers at the expense of double processing, I've been re-writing encoders to use a feature found in 'snprintf()'.  snprintf() returns the number of bytes that would have been written to the buffer if the buffer size (n) was large enough.  If n is zero, nothing is written, and the buffer may be a null pointer.

The feature can be used with a NULL apdu to compare with max size before using the actual buffer.  I completed this pattern in bacdcode.c module (although I had started with adding `apdu_offset` and that is not needed) and have been updating all the encoders as time permits.

The cov_encode_subscription() would benefit from this pattern, which usually only needs to change:

1. Change `&apdu[apdu_len]` to `apdu`
2. Add code to increment the `apdu` if `apdu` is not NULL after each encoding.
3. Update the Doxygen function header for `apdu`  so it includes `@param apdu  Pointer to the buffer for encoding, or NULL for only length`

Example:
```
    /* Recipient [0] BACnetRecipientProcess - opening */
    len = encode_opening_tag(apdu, 0);
    apdu_len += len;
    if (apdu) {
        apdu += len;
    }
```

The handler, which is passed `max_apdu` can be modified similar to bacnet_array_encode() pattern:
```
        len = encode_application_unsigned(NULL, array_size);
        if (len > max_apdu) {
            apdu_len = BACNET_STATUS_ABORT;
        } else {
            len = encode_application_unsigned(apdu, array_size);
            apdu_len = len;
        }
```

Is that something we should do resolve this change, or change it later?

metelik: I just saw your comment Well,.. Ad 1, 2 - yeah was thinking abou that but chose the least changes approach; Ad the pattern - we would need to rethink the encode_cov_suscription. As to when to do this I am indifferent the time may be the constraint here.

metelik: Secondly - how often a user is going to read the active cov subs ? Perhaps optimisation is not that critical.

skarg: The snprintf pattern is an overall attempt to reduce the security vulnerabilities using a simple pattern, so not really an optimization.  Someday maybe it will use the [C99 variable array size sprintfa() pattern](https://github.com/skarg/sprintfa).

skarg (approved):

PR #765, opened 2024-09-15
Bugfix/read property multiple client errors

skarg:


src/bacnet/basic/client/bac-rw.c
```diff
@@ -211,15 +211,29 @@ static void bacnet_read_property_ack_process(
...
+            return;
+        }
+        /* check for empty list */
+        if (rp_data->application_data_len == 0) {
+            bacapp_value_list_init(value, 1);
+            value->tag = BACNET_APPLICATION_TAG_EMPTYLIST;
+            rp_data->error_class = ERROR_CLASS_SERVICES;
+            rp_data->error_code = ERROR_CODE_SUCCESS;
+            if (bacnet_read_write_value_callback) {
+                bacnet_read_write_value_callback(device_id, rp_data, NULL);
```
nazaryev-cool, line 229: Shouldn't you pass `value` as a third parameter here?
Otherwise, it makes no sense to fill value->tag in the code above.

src/bacnet/rpm.c
```diff
@@ -689,27 +694,45 @@ void rpm_ack_object_property_process(
                 &rp_data->array_index);
             if (len <= 0) {
                 /* malformed */
```
nazaryev-cool, line 696: I've bumped into a false positive result here: when the code parses the last byte in the specific RPM response (last loop iteration, apdu[0] = 0x1F, apdu_len = 1, closing tag), it falsely recognizes it as a "malformation" and executes a callback with INVALID_TAG inside. 

Please check the [bip-simulator-rpm-invalid-tag.pcapng.gz](https://github.com/user-attachments/files/17006731/bip-simulator-rpm-invalid-tag.pcapng.gz) for a full message. For "real-time debugging" you can also try this BACnet Thermostat simulator (https://github.com/Kretiss/BACnet-simulator, `pip3 install bacpypes asyncore`, `python3 rooms_simulator.py -a 192.168.0.2/24 -p 47808`), and the issue will be easily reproduced with it if you do ReadPropertyMultiple on its object analog-input:10.

src/bacnet/rpm.c
```diff
@@ -739,24 +767,42 @@ void rpm_ack_object_property_process(
...
                 }
                 if (callback) {
                     callback(device_id, rp_data);
                 }
             }
         }
-        len = rpm_decode_object_end(apdu, apdu_len);
-        if (len) {
+        /*  list-of-results [1] SEQUENCE OF SEQUENCE */
+        if (bacnet_is_closing_tag_number(apdu, apdu_len, 1, &len)) {
```
nazaryev-cool, line 795: As far as I understand, you can reach here only if `apdu_len` = 0, as there is no other way to exit from the loop above (`while(apdu_len)`) rather than having `apdu_len` = 0.

PR #770, opened 2024-09-20
add ether support for mac os x

stargieg:
Depends on pcap (e.g. with Wireshark)
build with
#cmake .. -G"CodeBlocks - Unix Makefiles" -DBACDL_ETHER=ON -DBACDL_BIP=OFF -DBAC_ROUTING=OFF
run as normal user (not root) like Wireshark


stargieg:
new PR's for win32 and bsd

PR #774, opened 2024-09-22
Fix/port win32

stargieg:


skarg:
The pre-commit can be fixed by a) installing pre-commit and b) running pre-commit manually:
```
pre-commit run --all-files
```
Then commit the results.

skarg (approved):

PR #783, opened 2024-09-24
fix server app doesn't work with BAC_ROUTING

stargieg:


skarg:
Does this change change break the Gateway example?

stargieg:
> Does this change change break the Gateway example?
The gateway app is commented out in cmake.
I have enabled and compile with out err. But not tested so far.
I do this and report it here.

stargieg:
> Does this change change break the Gateway example?

Yes it is affected. Here is a new approach to the problem.
6a982396b48538315c1f5d714e035913802e7600


This line can be enabled. I have tested it with apple and linux.
https://github.com/bacnet-stack/bacnet-stack/blob/7b18add6d311d43eb61af405041080e6f66e110c/CMakeLists.txt#L777

skarg (approved):

PR #809, opened 2024-10-13
Fix the MSTP turnaround delay implementation.

nazaryev-cool:
The usleep() function takes microseconds, not milliseconds. So, T_turnaround needs to be calculated in microseconds. If we don't do this, at a baud rate of 9600, we end up waiting 4 microseconds instead of 4 milliseconds, which goes against the rule.


nazaryev-cool:
It might be a good idea to check mstp_port->SilenceTimer() and wait until it's greater than T_turnaround_ms.

P.S. From what I understand, this issue hasn't been noticed because modern USB-RS485 devices are slow enough that the delay happens naturally. That’s likely why it hasn't caused problems for anyone before.

skarg (approved):

PR #815, opened 2024-10-18
fix EXC_BAD_ACCESS in datalink_set() strcasecmp()

stargieg:
If -DBACDL_MULTIPLE and runtime env BACNET_DATALINK is null '''
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0) frame #0: 0x00007ff819a0e782 libsystem_c.dylib`strcasecmp_l + 91 '''

Set default datalink with priority bip,bip6,mstp,ethernet,arcnet


stargieg:
new PR folow

PR #853, opened 2024-11-12
Devel/object event state master

metelik:


src/bacnet/basic/object/bi.c
```diff
@@ -908,7 +908,9 @@ int Binary_Input_Read_Property(BACNET_READ_PROPERTY_DATA *rpdata)
         case PROP_STATUS_FLAGS:
             /* note: see the details in the standard on how to use these */
             bitstring_init(&bit_string);
-            bitstring_set_bit(&bit_string, STATUS_FLAG_IN_ALARM, false);
+            bitstring_set_bit(
+                &bit_string, STATUS_FLAG_IN_ALARM,
+                pObject->Event_State != EVENT_STATE_NORMAL);
```
skarg, line 913: please use  `Binary_Input_Event_State(rpdata->object_instance)` since pObject isn't guaranteed to be available (it is behind some `#ifdef`).

src/bacnet/basic/object/bv.c
```diff
@@ -915,7 +915,9 @@ int Binary_Value_Read_Property(BACNET_READ_PROPERTY_DATA *rpdata)
         case PROP_STATUS_FLAGS:
             /* note: see the details in the standard on how to use these */
             bitstring_init(&bit_string);
-            bitstring_set_bit(&bit_string, STATUS_FLAG_IN_ALARM, false);
+            bitstring_set_bit(
+                &bit_string, STATUS_FLAG_IN_ALARM,
+                pObject->Event_State != EVENT_STATE_NORMAL);
```
skarg, line 920: please use Binary_Value_Event_State(rpdata->object_instance) since pObject isn't guaranteed to be available (it is behind some #ifdef).

skarg (changes requested):

src/bacnet/basic/object/bi.c
metelik: Will do... May I ask why? the direct access through pointer is way more efficien...

skarg (approved):

PR #887, opened 2025-01-06
Devel/nc fd bbmd address choice 2

metelik:
Adding support for writing/reading of the FD_BBMD_Address property when its value is of the choice[2]:

BACnetHostAddress ::= CHOICE {
none [0] NULL,
ip-address [1] OCTET STRING, -- 4 octets for B/IP or 16 octets for B/IPv6
**name [2] CharacterString -- Internet host name (see RFC 1123)**
}
BACnetHostNPort ::= SEQUENCE {
host [0] BACnetHostAddress,
port


src/bacnet/basic/object/av.c
```diff
@@ -1438,6 +1438,341 @@ void Analog_Value_Intrinsic_Reporting(uint32_t object_instance)
 }
 
 #if defined(INTRINSIC_REPORTING)
+/**
```
metelik, line 1441: Not sure why these had not made to master already.... They're present in the av.h

test/bacnet/datalink/bvlc/src/main.c
```diff
@@ -859,14 +859,24 @@ static void test_BVLC_BBMD_Address(void)
     uint8_t apdu[480] = { 0 };
     int apdu_len = 0;
     int test_apdu_len = 0;
-    BACNET_IP_ADDRESS bbmd_address;
+    BACNET_HOST_N_PORT bbmd_address;
     BACNET_IP_ADDRESS test_bbmd_address;
     BACNET_ERROR_CODE error_code = ERROR_CODE_SUCCESS;
+#if 0
     bool status = false;
```
metelik, line 866: I didn't want to touch the bvlc_* api functions as they might be of use to someone - just not really of use here in this test.

metelik (commented):
Few comments on the code changes.

src/bacnet/basic/object/netport.c
```diff
@@ -12,6 +12,8 @@
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
+#include <ctype.h>
```
skarg, line 15: config.h is already included in bacdef.h so that it is the first define, and bacdef.h is designated as the first define. Please remove config.h here.

src/bacnet/basic/object/netport.c
```diff
@@ -1840,113 +1840,36 @@ bool Network_Port_BBMD_FD_Table_Set(uint32_t object_instance, void *fdt_head)
...
-    if (index < BACNET_NETWORK_PORTS_MAX) {
-        if (Object_List[index].Network_Type == PORT_TYPE_BIP) {
-            if ((Object_List[index].Network.IPv4.BBMD_IP_Address[0] != a) ||
-                (Object_List[index].Network.IPv4.BBMD_IP_Address[1] != b) ||
-                (Object_List[index].Network.IPv4.BBMD_IP_Address[2] != c) ||
-                (Object_List[index].Network.IPv4.BBMD_IP_Address[3] != d)) {
-                Object_List[index].Changes_Pending = true;
+        if (BBMD_Address) {
+            if (bbmd_address) {
+                *bbmd_address = *BBMD_Address;
```
skarg, line 1871: I prefer to use a function to copy structures.  See host_n_port_copy().

src/bacnet/basic/object/netport.c
```diff
@@ -1956,51 +1879,62 @@ bool Network_Port_Remote_BBMD_IP_Address_Set(
...
-    return value;
-}
+        if (Object_List[index].Network_Type == PORT_TYPE_BIP) {
+            dest_bbmd_address = &Object_List[index].Network.IPv4.BBMD_Address;
+        } else if (Object_List[index].Network_Type == PORT_TYPE_BIP6) {
+            dest_bbmd_address = &Object_List[index].Network.IPv6.BBMD_Address;
+        }
+
+        if (dest_bbmd_address) {
+            if ((dest_bbmd_address->host_name != bbmd_address->host_name) ||
```
skarg, line 1908: Can `host_n_port_same()` be used instead?

src/bacnet/basic/object/netport.c
```diff
@@ -3036,6 +2820,75 @@ bool Network_Port_IPv6_Gateway_Zone_Index_Set(
 }
 
 #if (BBMD_CLIENT_ENABLED) && defined(BACAPP_HOST_N_PORT)
+#define MAX_LABEL_LENGTH 63
+static bool is_valid_hostname(const BACNET_CHARACTER_STRING *const hostname)
```
skarg, line 2824: please move is_valid_hostname() into either hostnport.c or into bacstr.c (similar to utf8 check) and prefix appropriately.

src/bacnet/basic/object/netport.h
```diff
@@ -207,16 +207,11 @@ BACNET_STACK_EXPORT
 bool Network_Port_BBMD_FD_Table_Set(uint32_t object_instance, void *fdt_head);
 
 BACNET_STACK_EXPORT
-bool Network_Port_Remote_BBMD_IP_Address(
```
skarg, line 210: The IP address API is still valid for implementations that do not support DNS lookup, so we should not remove these in the header file.

src/bacnet/datalink/bvlc.c
```diff
@@ -2626,20 +2626,26 @@ const char *bvlc_result_code_name(uint16_t result_code)
...
+ * @param bbmd_address - HostNPort FD BBMD Address
  * @return length of the APDU buffer
  */
 int bvlc_foreign_device_bbmd_host_address_encode(
-    uint8_t *apdu, uint16_t apdu_size, const BACNET_IP_ADDRESS *ip_address)
+    uint8_t *apdu,
+    uint16_t apdu_size,
+    const BACNET_HOST_N_PORT *const bbmd_address)
 {
+#if (__STDC__) && (__STDC_VERSION__ >= 199901L)
```
skarg, line 2637: Maybe we should move 'port' to the beginning of struct BACnetHostNPort in order to use the simpler initialization and keep the code clean?

src/bacnet/datalink/bvlc6.c
```diff
@@ -1589,20 +1589,26 @@ int bvlc6_decode_distribute_broadcast_to_network(
...
+#endif
     int apdu_len = 0;
 
-    address.host_ip_address = true;
-    address.host_name = false;
-    octetstring_init(
-        &address.host.ip_address, &ip6_address->address[0], IP6_ADDRESS_MAX);
-    address.port = ip6_address->port;
+    if (bbmd_address) {
+        address = *bbmd_address;
```
skarg, line 1609: For structure copy, I prefer to use a function such as host_n_port_copy() which is designed for it.

src/bacnet/datalink/bvlc6.h
```diff
@@ -12,8 +12,10 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stddef.h>
+#include "bacnet/config.h"
```
skarg, line 15: config.h is included in bacdef.h as the first define is is only included there. Please remove config.h here.

test/bacnet/datalink/bvlc/src/main.c
skarg: We probably want to include them to have better Line of Code Coverage (see lcov in Cmake test builds).

src/bacnet/basic/object/netport.c
metelik: done.

src/bacnet/basic/object/netport.h
metelik: There are two things into it: #1. It is fetchable through Network_Port_Remote_BBMD_Address that can be Null(0), IP(1) or HostName[2] #2. Background: “e.g” 135.1 – 7.3.2.46.1.1 Clause J.1.2.1 requires that “B/IP devices shall be capable of resolving internet host names to IP addresses”. I can share the full exchange with the BTL labs....

src/bacnet/basic/object/netport.c
metelik: Will do.

src/bacnet/datalink/bvlc.c
metelik: In this case I like the c11 and later approach. Perhaps we could keep the differentiation between standards and some day the c90 would become a relict....

src/bacnet/datalink/bvlc6.h
metelik: Hmmmm.... It was missing the HOST_N_PORT datatype - shall take a second look. Thank you for the hint.

src/bacnet/datalink/bvlc.c
skarg: Although c11 declaration initializer is more useful in some cases, when we are simply setting all the values to zero or false it is a lot easier to maintain using the ` = { 0 };` initializer.  If the `0` is giving a compiler warning, we can just shift a numeric member to the beginning of the structure.

src/bacnet/basic/object/netport.c
metelik: done.

src/bacnet/basic/object/netport.c
metelik: done.

test/bacnet/datalink/bvlc/src/main.c
metelik: restored = done.

src/bacnet/datalink/bvlc.c
metelik: Nope = { 0 } works for all. Yet I prefer the c99+ structs init. Besides we'd have the code amelioration at hand when the time comes to scrap c90 compatibility. I works well w/ c99, c11, c17 it may not be the prettiest thing in the world the #if but I wouldn't like to change this part.

src/bacnet/datalink/bvlc6.c
metelik: done.

src/bacnet/datalink/bvlc6.h
metelik: done.

src/bacnet/basic/object/netport.h
metelik: restored.

src/bacnet/basic/object/netport.c
metelik: done.

metelik (commented):
Re-worked

src/bacnet/basic/object/netport.c
```diff
@@ -2851,6 +2885,88 @@ bool Network_Port_IPv6_DHCP_Server_Set(
...
+    if (index < BACNET_NETWORK_PORTS_MAX) {
+        if (Object_List[index].Network_Type == PORT_TYPE_BIP6) {
+#if (__STDC__) && (__STDC_VERSION__ >= 199901L)
+            struct timeval tv = { .tv_sec = 0, .tv_usec = 0 };
+#else
+            struct timeval tv = { 0, 0 };
+#endif
+
+            Object_List[index].Network.IPv6.IP_DHCP_Lease_Time = value;
+            if (gettimeofday(&tv, NULL) != 0) {
```
skarg, line 2934: please use mstimer for the DHCP lease time tracking in this basic object.  I don't think datetime_local() - calendar time - is the right time in seconds to use here.

src/bacnet/basic/object/netport.c
```diff
@@ -31,6 +32,15 @@
 #include "bacnet/datalink/bvlc6.h"
 #endif
 
+#if !defined(_MSC_VER)
+#include <sys/time.h>
+#else
+#include <winsock.h>
+
+/* defined in ports/win32/datetime-init.c */
+int gettimeofday(struct timeval *tp, void *tzp);
```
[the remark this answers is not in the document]
metelik: I initially looked at that function but it would be enormously awkward to use it when we need the seconds from epoch and not necessarily Tuesday 27th of March 2025 11:17 p.m. <-- absolutely dreadful to convert.

skarg:
Added handling for the host name option of BACnetHostNPort in [#997]

PR #894, opened 2025-01-20
bacdcode.c  : Don't allow two functions to add a neg number for the l…

jd-imi:
…ength

The functions are encode_application_date and encode_context_date


skarg:
I changed encode_application_date and encode_context_date in a different way to fix the problem that existed - see [#897](https://github.com/bacnet-stack/bacnet-stack/pull/897)

PR #898, opened 2025-01-29
Fixed IPv6 handler to ignore original-broadcast when registered as foreign-device

skarg:


metelik (approved):
I'd do the if (0 != Remote_BBMD.port) instead of idiom if (Remote_BBMD.port)  but am indifferent between the two :)

PR #920, opened 2025-02-20
Added instance check to Device_Object_Name_Copy

Vimalaramani-U:
In device.c - Device_Object_Name_Copy function, instance number check is missing before calling function for fetching the Object_Name. This was used by handler_who_has.

Added a condition to check whether the object_instance is valid then call Object_Name for valid object_instance.


skarg:
I cleaned up the code using pre-commit, and added the changes to all the Device_Object_Name_Copy() functions in all the device.c modules in the library. See #922 for details.

PR #956, opened 2025-03-31
Added write present value callbacks for Analong and Integer values

dmitttri:
This PR adds write present value callbacks to Analog and Integer value basic objects

### DIFF
--- a/src/bacnet/basic/object/av.c
+++ b/src/bacnet/basic/object/av.c
@@ -32,6 +32,9 @@
 static OS_Keylist Object_List;
 /* common object type */
 static const BACNET_OBJECT_TYPE Object_Type = OBJECT_ANALOG_VALUE;
+/* callback for present value writes */
+static analog_value_write_present_value_callback
+    Analog_Value_Write_Present_Value_Callback;
 
 /* clang-format off */
 /* These three arrays are used by the ReadPropertyMultiple handler */
@@ -922,6 +925,7 @@ bool Analog_Value_Write_Property(BACNET_WRITE_PROPERTY_DATA *wp_data)
 {
     bool status = false; /* return value */
     int len = 0;
+    float old_value = 0.0f;
     BACNET_APPLICATION_DATA_VALUE value = { 0 };
     ANALOG_VALUE_DESCR *CurrentAV;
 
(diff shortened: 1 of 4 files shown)


skarg:
I enabled the github workflow checks. Pre-commit is currently failing (style).  To fix, run `pre-commit run --all-files` (usually on a POSIX system) and commit the fixed files. 

Note: there is a `configure` script in the root folder for Debian/Ubuntu which can also be used to install the build and checking tools. The .github/workflows/ folder has the instructions that the checks use to do their checks, usually from a recipe in the root Makefile.

src/bacnet/basic/object/av.c
```diff
@@ -962,13 +966,20 @@ bool Analog_Value_Write_Property(BACNET_WRITE_PROPERTY_DATA *wp_data)
...
                                wp_data->object_instance, value.type.Real,
                                wp_data->priority)) {
-                    status = true;
-                } else {
-                    wp_data->error_class = ERROR_CLASS_PROPERTY;
-                    wp_data->error_code = ERROR_CODE_VALUE_OUT_OF_RANGE;
+                        status = true;
+                        if (Analog_Value_Write_Present_Value_Callback) {
+                            Analog_Value_Write_Present_Value_Callback(
+                                wp_data->object_instance, old_value, value.type.Real);
```
skarg, line 977: Although in this av.c implementation there is no priority-array, this section of code should not rely on that knowledge.  Please change value.type.Real to a call to Analog_Value_Present_Value() so that if the priority was used, it would read the correct present-value.

src/bacnet/basic/object/iv.c
```diff
@@ -554,9 +558,14 @@ bool Integer_Value_Write_Property(BACNET_WRITE_PROPERTY_DATA *wp_data)
             status = write_property_type_valid(
                 wp_data, &value, BACNET_APPLICATION_TAG_SIGNED_INT);
             if (status) {
+                old_value = Integer_Value_Present_Value(wp_data->object_instance);
                 Integer_Value_Present_Value_Set(
                     wp_data->object_instance, value.type.Signed_Int,
                     wp_data->priority);
+                if (Integer_Value_Write_Present_Value_Callback) {
+                    Integer_Value_Write_Present_Value_Callback(
+                        wp_data->object_instance, old_value, value.type.Signed_Int);
```
skarg, line 567: Although in this av.c implementation there is no priority-array, this section of code should not rely on that knowledge. Please change value.type.Signed_Int to another call to Integer_Value_Present_Value() so that if the priority was used, it would read the correct present-value.

dmitttri: Done

skarg (approved):

PR #958, opened 2025-04-03
specify a target for discovery

ryanjmulder:
This allows folks to use something like `bacdiscover --mac "192.168.1.79:47808"` to discover a BACnet server running on a different port on the same host (as long as the server is using `handler_who_is_unicast`), which is at least useful for developing a BACnet server, but probably has other uses.

- added `bacnet_discover_dest_set` to `bac-discover`
- copied `--dadr` and `--mac` support from `readprop` to `server-discover`

### DIFF
--- a/apps/server-discover/main.c
+++ b/apps/server-discover/main.c
@@ -266,6 +266,7 @@ int main(int argc, char *argv[])
                 if ((long_value >= 0) &&
                     (long_value <= BACNET_BROADCAST_NETWORK)) {
                     dnet = (uint16_t)long_value;
+                    specific_address = true;
                 }
             }
         } else if (strcmp(argv[argi], "--mac") == 0) {
--- a/apps/server-discover/main.c
+++ b/apps/server-discover/main.c
@@ -301,34 +301,6 @@ int main(int argc, char *argv[])
     }
     if (specific_address) {
         bacnet_address_init(&dest, &mac, dnet, &adr);
-        if (adr.len && mac.len) {
-            memcpy(&dest.mac[0], &mac.adr[0], mac.len);
-            dest.mac_len = mac.len;
-            memcpy(&dest.adr[0], &adr.adr[0], adr.len);
-            dest.len = adr.len;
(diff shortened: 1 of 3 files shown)


apps/server-discover/main.c
```diff
@@ -262,6 +299,37 @@ int main(int argc, char *argv[])
             BACNET_MAX_INSTANCE);
         return 1;
     }
+    if (specific_address) {
+        bacnet_address_init(&dest, &mac, dnet, &adr);
+        if (adr.len && mac.len) {
```
skarg, line 304: It seems that apps/readprop has some cruft laying around.  The bacnet_address_init() does (mostly) everything that the set of if/else is doing. [apps/writeprop](https://github.com/bacnet-stack/bacnet-stack/blob/31af2507fbe201029d438cb06d84c4e6d364ebed/apps/writeprop/main.c#L457) and a few other apps have been updated. 

Here is what this section should look like:
```
        if (specific_address) {
            bacnet_address_init(&dest, &mac, dnet, &adr);
        }
```

apps/server-discover/main.c
```diff
@@ -262,6 +299,9 @@ int main(int argc, char *argv[])
             BACNET_MAX_INSTANCE);
         return 1;
     }
+    if (specific_address) {
+        bacnet_address_init(&dest, &mac, dnet, &adr);
```
skarg, line 303: dnet could be -1 here. please add a check for dnet in range and assign to broadcast if out of range by default.  for example:
```
        if ((dnet < 0) || (dnet > BACNET_BROADCAST_NETWORK)) {
            dnet = BACNET_BROADCAST_NETWORK;
        }
        bacnet_address_init(&dest, &mac, dnet, &adr);
```

ryanjmulder: `--dnet` actually had two branches in the command line parsing loop, so I removed one.

The other was already bounds-checking the result of `strtol`, which allows `dnet` to be initialized with a valid default of `BACNET_BROADCAST_NETWORK` so the bounds check doesn't need to be repeated at point of use.

skarg (approved):

PR #981, opened 2025-04-29
Fixed the ghost Device ID 0 in the I-Am response when the actual rout…

skarg:
…ed devices are less than the MAX_NUM_DEVICES.


jshenguru:
To fix this issue, all occurrence of `MAX_NUM_DEVICES` in in the _gw_device.c_ should be replaced with `Num_Managed_Devices` except for that in `DEVICE_OBJECT_DATA Devices[MAX_NUM_DEVICES];` and  that in the `Add_Routed_Device()` function.
I have made those changes in my local copy of source code and proved the fix works as expected.

PR #1007, opened 2025-05-28
Linux MS/TP 76800

ryanjmulder:
Fixes MS/TP 76800 on Linux

### Background

BACnet MS/TP has 6 standard baud rates (`9.2.3 Timing`)
POSIX systems have [16 standard baud rates](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/termios.h.html) but the intersection of the two sets is missing `57600`, `76800`, and `115200`.
glibc's Linux support includes [extra non-posix baud rates](https://codebrowser.dev/glibc/glibc/sysdeps/unix/sysv/linux/bits/termios-baud.h.html) which covers  `57600` and `115200`.

`76800` then, needed special handling.
Currently, `master` uses an old Linux hack that lets a divisor be applied when the baud is set to `B38400`, and that hack does not work with the [Linux ACM driver](cdc_acm). It fails with error `RS485 bad custom divisor 0, base baud 0`. (`master` does work with `ftdi_sio` driver)

[Linux 2.6.20 (released in 2007)](https://kernelnewbies.org/Linux_2_6_20#Various_core_changes) added a new interface (`termios2`) to set the baud rate directly with an integer (search for `TTY: termios revamp, adds proper speed control`). 

glibc has been [discussing Linux's baud rate interface](https://sourceware.org/bugzilla/show_bug.cgi?id=10339) since 2009, but they have made some progress this year.

So, the solution to universal Linux support for `76800` is to use the `termios2` interface [directly via ioctl](https://man7.org/linux/man-pages/man2/TCSETS.2const.html).

This works with both `cdc_acm` and `ftdi_sio` and the uart built into my Apollo Lake generation Intel SoC

### Changes

* add `ports/linux/termios2.c/h` with `termios2` implementations of `tcsetattr`, etc.
* use `termios2` in `ports/linux/rs485` and `ports/linux/dlmstp_port`
* add `RS485_Set_Config` to `ports/linux/rs485` to opt-in to [automatic use of RTS to control direction](https://docs.kernel.org/driver-api/serial/serial-rs485.html)
* removed duplicate code in `RS485_Send_Frame` and `RS485_Check_UART_Data`
* replace `bzero` with `memset` because [bzero was removed in POSIX.1-2008](https://man7.org/linux/man-pages/man3/bzero.3.html)
* remove direct assignment of `shared_port_data.RS485_Baud` from `apps/router/mstpmodule.c` because it already used `dlmstp_set_baud_rate` instead

### DIFF
--- a/apps/fuzz-afl/Makefile
+++ b/apps/fuzz-afl/Makefile
@@ -24,10 +24,6 @@ PORT_MSTP_SRC = \
 	$(BACNET_SRC_DIR)/bacnet/datalink/mstptext.c \
 	$(BACNET_SRC_DIR)/bacnet/datalink/crc.c
 
-ifeq (${BACNET_PORT},linux)
-	PORT_MSTP_SRC += $(BACNET_PORT_DIR)/termios2.c
-endif
-
 PORT_BIP_SRC = \
 	$(BACNET_PORT_DIR)/bip-init.c \
 	$(BACNET_SRC_DIR)/bacnet/datalink/bvlc.c \
--- a/apps/fuzz-libfuzzer/Makefile
+++ b/apps/fuzz-libfuzzer/Makefile
@@ -26,10 +26,6 @@ PORT_MSTP_SRC = \
 	$(BACNET_SRC_DIR)/bacnet/datalink/mstptext.c \
 	$(BACNET_SRC_DIR)/bacnet/datalink/crc.c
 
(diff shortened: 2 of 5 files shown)


skarg (approved):

ryanjmulder:
> The Makefile build (default) is broken.

Makefile build is fixed, thanks @skarg

PR #1025, opened 2025-06-18
enable writing mstp mac address and link speed via network_port

ryanjmulder:
also implement Device_Reinitialize_State_Set

### DIFF
--- a/src/bacnet/basic/object/netport.c
+++ b/src/bacnet/basic/object/netport.c
@@ -125,6 +125,13 @@ struct object_data {
 
 static struct object_data Object_List[BACNET_NETWORK_PORTS_MAX];
 
+#ifndef MSTP_LINK_SPEED_LIST
+#define MSTP_LINK_SPEED_LIST 9600, 19200, 38400, 57600, 76800, 115200
+#endif
+static unsigned MSTP_LINK_SPEEDS[] = { MSTP_LINK_SPEED_LIST };
+static unsigned MSTP_LINK_SPEEDS_MAX =
+    sizeof MSTP_LINK_SPEEDS / sizeof(unsigned);
+
 /* These three arrays are used by the ReadPropertyMultiple handler */
 static const int Network_Port_Properties_Required[] = {
     /* unordered list of required properties */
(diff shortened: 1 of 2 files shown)


src/bacnet/basic/object/netport.c
```diff
@@ -125,6 +125,13 @@ struct object_data {
 
 static struct object_data Object_List[BACNET_NETWORK_PORTS_MAX];
 
+#ifndef MSTP_LINK_SPEED_LIST
```
skarg: Please use a list of integers (array) terminated with `-1`, similar to property lists in all the objects, rather than a macro, for the list of link speeds.  The integer list is a more familiar pattern that is already used in the basic objects.

ryanjmulder: done!
Although, I kept the calculated constant `MSTP_Port_Link_Speeds_Array_Size`  for use with `bacnet_array_encode`

src/bacnet/basic/object/netport.c
```diff
@@ -170,10 +170,17 @@ static const int MSTP_Port_Properties_Optional[] = {
     PROP_NETWORK_NUMBER,
     PROP_NETWORK_NUMBER_QUALITY,
     PROP_LINK_SPEED,
+    PROP_LINK_SPEEDS,
 #endif
     -1
 };
 
+static const int MSTP_Port_Link_Speeds[] = {
```
skarg: PROP_LINK_SPEEDS and its API are already implemented in [ports/stm32f4xx](https://github.com/bacnet-stack/bacnet-stack/blob/master/ports/stm32f4xx/netport.c) and [ports/stm32f10x](https://github.com/bacnet-stack/bacnet-stack/blob/master/ports/stm32f10x/netport.c) and [ports/bdk-atxx4-mstp](https://github.com/bacnet-stack/bacnet-stack/blob/master/ports/bdk-atxx4-mstp/netport.c) - can you port one of their implementations to this module?

ryanjmulder: I copied the Link_Speeds implementation from bdk-atxx4-mstp.

However, I left the validation that the `Link_Speed` is in `Link_Speeds` in `Network_Port_Write_Property`, as opposed to `Network_Port_Link_Speed_Set`, because `Network_Port_Link_Speed_Set` should be able to set the `Link_Speed` to `0`, but `Network_Port_Write_Property` should not.

`Network_Port_Write_Property` also still checks that the port type is MSTP

skarg (approved):

PR #1031, opened 2025-06-25
feat(ports/linux): Allow specifying bind IP address via environment variable

huduo0812:
### Description:

**Motivation and Problem**

On Linux systems, when a network interface is configured with multiple IP addresses (e.g., using IP aliasing), the current implementation using `ioctl` with `SIOCGIFADDR` only retrieves the primary IP address of the interface. This limitation makes it impossible to bind the BACnet service to a specific secondary or aliased IP address, which is a necessary feature for applications running in complex or multi-homed network environments.

**Solution**

This pull request introduces a new environment variable, `BACNET_BIND_ADDRESS`, to provide more granular control over the IP address binding process on Linux.

The logic is as follows:
1.  The application first checks for the `BACNET_BIND_ADDRESS` environment variable.
2.  If the variable is set and contains a valid IPv4 address string (e.g., "192.168.1.200"), the BACnet/IP data link layer will use this address for the socket `bind()` call.
3.  If the environment variable is not set or its value is invalid, the application gracefully falls back to the original behavior of auto-detecting the interface's primary IP address.

This change is fully backward-compatible and provides essential flexibility for users who need to operate on specific IP addresses without altering the default behavior for other users.

**How to Use**

To bind to a specific IP address, set the environment variable before running a BACnet application.

**Example:**
```bash
# Set the desired IP address for binding
export BACNET_BIND_ADDRESS="192.168.1.200"

# Run a BACnet application, which will now bind to 192.168.1.200
./bin/bacwi 12345

### DIFF
--- a/ports/linux/bip-init.c
+++ b/ports/linux/bip-init.c
@@ -11,6 +11,7 @@
 #include <net/if.h>
 #include <sys/socket.h>
 #include <sys/ioctl.h>
+#include <arpa/inet.h>
 #include <linux/netlink.h>
 #include <linux/rtnetlink.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by skarg** [comment]
The Windows port/win32 uses a network address rather than an interface, but reuses the same environment variable `BACNET_IFACE=192.168.1.200`.  Is it possible to do the same with Linux or BSD?

Maybe using a test like:
```
static int is_valid_ipv4(const char *ifname) {
    struct sockaddr_in sa;
    // inet_pton returns 1 for success, 0 for invalid, -1 for error
    return inet_pton(AF_INET, ifname, &(sa.sin_addr)) == 1;
}
```

I understand that it is overloading the usage of BACNET_IFACE, but the precedent on Windows has been in the library for a long time.

huduo0812:
> The Windows port/win32 uses a network address rather than an interface, but reuses the same environment variable `BACNET_IFACE=192.168.1.200`. Is it possible to do the same with Linux or BSD?
> 
> Maybe using a test like:
> 
> ```
> static int is_valid_ipv4(const char *ifname) {
>     struct sockaddr_in sa;
>     // inet_pton returns 1 for success, 0 for invalid, -1 for error
>     return inet_pton(AF_INET, ifname, &(sa.sin_addr)) == 1;
> }
> ```
> 
> I understand that it is overloading the usage of BACNET_IFACE, but the precedent on Windows has been in the library for a long time.

Thank you very much for your reply.

Regarding your suggestion to use an IP address for the `BACNET_IFACE` variable on Linux, I can confirm that this does not seem to work on my system (Ubuntu 18.04.6 LTS).

My specific setup was as follows:
1.  The BACnet device I needed to access was at `192.168.1.199`.
2.  To communicate with it, I added the IP address `192.168.1.200` to my local network card.
image

3.  I then set the environment variable: `export BACNET_IFACE=192.168.1.200`.


When I tried to run the demo program with this configuration, it produced an error and failed to start. It seems that, unlike the Windows port, the Linux implementation expects an interface name (e.g., `eth0`) rather than an IP address.
image


I have already worked around the problem I was facing. My colleague changed the IP of the target device to `192.168.11.199`, which resolved my connectivity issue.

Thank you again for your help and the discussion.

skarg:
No changes were need so I am closing the code review request.

PR #1036, opened 2025-07-01
Avoid extra socket if the unicast and broadcast address are the same

skarg:
Avoid extra socket if the unicast and broadcast address are the same. 
This happens with VPN tunnels or P2P connections.

### DIFF
--- a/ports/bsd/bip-init.c
+++ b/ports/bsd/bip-init.c
@@ -36,6 +36,8 @@ static struct in_addr BIP_Broadcast_Binding_Address;
 static bool BIP_Debug = false;
 /* interface name */
 static char BIP_Interface_Name[IF_NAMESIZE] = { 0 };
+/* point-to-point interface flag */
+static bool BIP_Point_To_Point = false;
 
 /**
  * @brief Print the IPv4 address with debug info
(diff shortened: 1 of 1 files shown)


skarg:
The AI generated code is nonsensical to me, so I am declining.  The VPN changes have already been made to ports/bsd to detect that the IP addresses for unicast and broadcast are the same.

bkoldun:
> Steve Karg

> The AI generated code is nonsensical to me, so I am declining. The VPN changes have already been made to ports/bsd to detect that the IP addresses for unicast and broadcast are the same.

The VPN changes don't work with OpenVPN. While they do work with WireGuard, WireGuard isn't sufficient for our needs.

bkoldun:
> Yes, I'd be happy to fix with a simpler and less invasive change that I can understand.

@skarg Could you please review the few lines that fix the issue? I believe it’s clearer now. I’ve attached the full updated `ports/bsd/bip-init.c` file as `bip-init.txt` below, and you can see the changed lines in the screenshot.

Знімок екрана 2025-08-05 о 23 21 46

[bip-init.txt](https://github.com/user-attachments/files/21606654/bip-init.txt)

bkoldun:
PR https://github.com/bacnet-stack/bacnet-stack/pull/1062

PR #1061, opened 2025-08-05
Added fixups to Microsoft Visual Studio build: added server-mini, etc.

BACnetEd:
Added fixups to Microsoft Visual Studio build: added server-mini, mstpcap readprop, writeprop projects. Tested server, server-mini, mstpcap


apps/mstpcap/mstpcap.txt
```diff
@@ -149,13 +149,18 @@ to be 300 baud.
...
+Worked just fine capturing at 38400 baud.
+
 Linux (used with Debian Lenny and Fedora 15)
 http://www.connecttech.com/KnowledgeDatabase/kdb309.htm
 As root:
 Change USB so I can use it later as normal user:
 # chmod 777 /dev/ttyUSB0  -
 Print current info about the device:
-# setserial /dev/ttyUSB0 �a
+# setserial /dev/ttyUSB0 –a
```
skarg: please use simple ASCII in txt files

apps/mstpcap/mstpcap.txt
```diff
@@ -166,7 +171,7 @@ Make custom speed:
 24000000/312 gives 76923 baudrate:
 # setserial /dev/ttyUSB0 divisor 312
 Print to make sure changes got applied:
-# setserial /dev/ttyUSB0 �a
+# setserial /dev/ttyUSB0 –a
```
skarg: please use simple ASCII in txt files

apps/mstpcap/mstpcap.txt
```diff
@@ -149,13 +149,18 @@ to be 300 baud.
 
 So to capture at 76800 baud type: mstpcap.exe COM2 300
 
+Windows 11 (from edward@bac-test.com)
+Using Waveshare USB to RS485 converter on COM4: https://www.waveshare.com/product/usb-to-rs485-b.htm
```
skarg: please adjust lines within 80 columns, where possible

src/bacnet/datalink/dlenv.c
```diff
@@ -31,7 +31,9 @@
 #include "bacnet/datalink/datalink.h"
 #include "bacnet/datalink/dlenv.h"
 #include "bacnet/datalink/dlmstp.h"
+#ifndef _MSC_VER
```
skarg: Maybe we should use `#ifdef BACFILE` here if no file object is needed for the application?
`_MSC_VER` should only be used where there are compiler differences, and hopefully only happens in platform.h file.

skarg (changes requested):

apps/mstpcap/mstpcap.txt
BACnetEd: I think you mixed up pre- and post- change here... line 163 is the current..

skarg: The [Files Changed](https://github.com/bacnet-stack/bacnet-stack/pull/1061/files) view shows it the other way around. Hmm...

ports/win32/rs485.c
```diff
@@ -252,12 +252,28 @@ static void RS485_Cleanup(void)
...
-        RS485_Port_Name, GENERIC_READ | GENERIC_WRITE, 0, 0, OPEN_EXISTING,
-        /*FILE_FLAG_OVERLAPPED */ 0, 0);
+    // Convert RS485_Port_Name (char*) to wide string (wchar_t*) for CreateFileW
+    wchar_t wRS485_Port_Name[256];
+    size_t converted = 0;
+    mbstowcs_s(
+        &converted, wRS485_Port_Name,
+        sizeof(wRS485_Port_Name) / sizeof(wchar_t), RS485_Port_Name, _TRUNCATE);
+
+    RS485_Handle = CreateFileW(
```
skarg: @BACnetEd Rather than jump through hoops to get a wide character, can we simply change `CreateFile` to `CreateFileA`?

CreateFile a macro that resolves to either CreateFileA or CreateFileW at compile time. The resolution depends on whether the UNICODE preprocessor macro is defined in your project settings.

skarg: I changed all the win32 usage of CreateFile to CreateFileA in [#1076].  Can you see if that works for you?

PR #1062, opened 2025-08-06
fix(ports/bsd): Use BBMD for broadcast on VPN interfaces

bkoldun:


skarg:
The BBMD Foreign Device Registration manages the broadcast address in h_bbmd.c module to convert to a unicast to the BBMD.

1. dlenv.c -> calls bvlc_register_with_bbmd() when getenv("BACNET_BBMD_ADDRESS"); address is stored in Remote_BBMD

2. h_bbmd.c -> bvlc_send_pdu() checks Remote_BBMD.port to be non-zero for BACNET_BROADCAST_NETWORK or mac_len == 0 and sends bvlc_encode_distribute_broadcast_to_network().

This duplicates the changes you are proposing.  Is your application missing some of those function calls?

PR #1104, opened 2025-09-20
ai.c used a uint8_t for units (needs to be 16 bits)

BACnetEd:
ai.c was using a uint8_t for the units and this resulted in corruptions. I changed all uses of "units" to "BACNET_ENGINEERING_UNITS and in doing so found a few other misplaced casts.


src/bacnet/version.h
```diff
@@ -15,8 +15,8 @@
 #define BACNET_VERSION(x, y, z) (((x) << 16) + ((y) << 8) + (z))
 #endif
 
-#define BACNET_VERSION_TEXT "1.4.1"
-#define BACNET_VERSION_CODE BACNET_VERSION(1, 4, 1)
+#define BACNET_VERSION_TEXT "1.4.2"
```
skarg: We should change the version in a separate Pull Request by changing the version and updating CHANGELOG.txt.

BACnetEd: Hi Steve. OK, finding my feet here. 'Trivial changes' without spinning ver in future.

PR #1148, opened 2025-11-18
Added segmentation support for server reply. 

skarg:
Added segmentation support for server devices for some services. Configure BACNET_SEGMENTATION_ENABLED=1 to include in the library, and adjust BACNET_MAX_SEGMENTS_ACCEPTED for maximum number of segments.


Sekiro-kost:
Hello,

What is the status of this addition? Have you encountered any difficulties? It is something I need for a major project of my own.
Thanks in advance.

skarg:
The patch, based on an earlier segmentation branch, was working. I had merged it in at one point, but started seeing problems with non segmented and reverted it. I think those problems have since been resolved, but I haven't tested in for some time. I have tried to keep the branch up to date and also to merge into master any code that can exist on its own.

The only problem with this patch at the moment (other than being incomplete for both client and server and all the services and examples) is the dynamic memory blob handling where I would prefer to use a chained packet pool instead for storing segments of messages.

You are welcome to test it and let us know or update the branch with fixes.

PR #1155, opened 2025-11-20
linux: fix subnet prefix calculation

SyedAmerGilani:
previous implementation always returned 0.


skarg:
We use 'pre-commit' tool to check consistent style and spelling. Can you run it an re-commit?

PR #1299, opened 2026-04-10
Fix stack OOB read in decode_tag_number_and_value()

JorgeBarredo14:
## Summary

`decode_tag_number_and_value()` in `src/bacnet/bacdcode.c` reads up to 5 bytes past the caller buffer when the first APDU byte has `IS_EXTENDED_VALUE` set, because the function takes no buffer-length parameter.

## Root cause

Lines 695-712: when `IS_EXTENDED_VALUE(apdu[0])` is true, the function unconditionally reads `apdu[1]`. If that byte equals `0xFF`, it calls `decode_unsigned32` which reads four more bytes -- all without any length check.

## Fix

Add an early-return guard before `decode_tag_number()` that rejects extended-value tags with `BACNET_STATUS_ERROR`. The existing code block becomes unreachable and can be removed in a follow-up cleanup. The length-aware variant `bacnet_tag_number_and_value_decode()` is unaffected.

## Metadata

- **CWE**: CWE-125 (Out-of-bounds Read)
- **Severity**: Moderate (CVSS 5.3)
- **Advisory**: GHSA-cr77-f6f9-494h
- **Found during**: academic security research

### DIFF
--- a/src/bacnet/bacdcode.c
+++ b/src/bacnet/bacdcode.c
@@ -692,6 +692,20 @@ int decode_tag_number_and_value(
     uint16_t value16;
     uint32_t value32;
 
+    /* This legacy function cannot validate buffer bounds because
+     * it lacks an apdu_size parameter. Reject extended-value tags
+     * to prevent out-of-bounds reads. New code should use
+     * bacnet_tag_number_and_value_decode() instead. */
+    if (IS_EXTENDED_VALUE(apdu[0])) {
+        if (tag_number) {
+            *tag_number = (uint8_t)(apdu[0] >> 4);
+        }
+        if (value) {
+            *value = 0;
+        }
+        return BACNET_STATUS_ERROR;
+    }
+
     len = decode_tag_number(&apdu[0], tag_number);
     if (IS_EXTENDED_VALUE(apdu[0])) {
         /* tagged as uint32_t */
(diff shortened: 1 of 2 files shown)


skarg (approved):

JorgeBarredo14:
Hi Steve,

The CI `unittest` failure (exit code 8 = 8 tests failed) is expected with this change.
The early-return guard rejects **all** extended-value tags, which breaks existing tests
that call `decode_tag_number_and_value()` with valid extended-value inputs in controlled
(safe) contexts.

There are two approaches to fix this:

### Approach A — Update the tests (recommended)
Migrate the 8 failing test call-sites to use `bacnet_tag_number_and_value_decode()` 
(the length-aware variant). This is the safest long-term solution since the legacy
function is inherently unsafe for untrusted input. Example diff for one test:

```c
 // Before (unsafe, no bounds check):
 len = decode_tag_number_and_value(&apdu[0], &tag_number, &value);
 
 // After (safe, bounds-checked):
 len = bacnet_tag_number_and_value_decode(&apdu[0], sizeof(apdu), &tag_number, &value);
```

I'm happy to push this update to the PR if you prefer this approach.

Approach B — Narrow the guard
Instead of rejecting all extended-value tags, only return an error when the function
is called from external/untrusted paths (though this is hard to enforce at the function
level without a length parameter).

Let me know which direction you'd prefer and I'll update the PR accordingly.

skarg:
Use approach A: Migrate the 8 failing test call-sites to use bacnet_tag_number_and_value_decode()
(the length-aware variant).

JorgeBarredo14:
Pushed a follow-up commit migrating the 3 unguarded test call-sites to bacnet_tag_number_and_value_decode() as discussed. CI should pass now.

skarg:
Removing the deprecated functions is the better solution. Closing with plans to remove them.

PR #1320, opened 2026-04-29
Support routed virtual device backup/restore reinitialization

kim-com:
## Summary

Related to #1287.

This updates routed device reinitialization handling so virtual devices can participate in Backup/Restore flows without changing the existing gateway-device behavior.

## Key changes

- Keep existing behavior when `BAC_ROUTING` is disabled.
- Keep existing gateway-device behavior when `BAC_ROUTING` is enabled.
- Allow routed virtual devices to receive ReinitializeDevice service requests, but only handle Backup/Restore related states.
- Return `ERROR_CLASS_SERVICES` / `ERROR_CODE_OPTIONAL_FUNCTIONALITY_NOT_SUPPORTED` for routed virtual-device `COLDSTART`, `WARMSTART`, and `ACTIVATE_CHANGES` instead of sending a Simple ACK.
- Keep DCC blocked for routed virtual devices.
- Keep `Reinit_Password` as `const char *` with the existing default value.
- Add per-device routing state only under `BAC_ROUTING`.
- Add Backup/Restore fields only under `BAC_ROUTING && BACNET_BACKUP_RESTORE`.

## Tests

Added device test variants that reuse the same test source across the macro matrix:

- `test_device`: `BAC_ROUTING` off, `BACNET_BACKUP_RESTORE` off
- `test_device_backup_restore`: `BAC_ROUTING` off, `BACNET_BACKUP_RESTORE` on
- `test_device_bac_routing`: `BAC_ROUTING` on, `BACNET_BACKUP_RESTORE` off
- `test_device_bac_routing_backup_restore`: `BAC_ROUTING` on, `BACNET_BACKUP_RESTORE` on

Additional coverage verifies that routed virtual devices reject unsupported ReinitializeDevice states:

- `COLDSTART`
- `WARMSTART`
- `ACTIVATE_CHANGES`

Verification performed:

- Built and ran all four device test variants successfully.
- Ran the full top-level CTest suite successfully: `274/274` tests passed.

## Follow-up note

The device-object handler direction in `feature/refactor-device-object-handling` looks useful for a future refactor, but this PR intentionally keeps that broader `device_h` / per-device object-handler restructuring out of scope. The current change keeps the Issue #1287 behavior focused and limited to routed device Backup/Restore handling plus the related unit-test matrix.

### DIFF
--- a/src/bacnet/basic/object/device.c
+++ b/src/bacnet/basic/object/device.c
@@ -1460,17 +1460,14 @@ static uint32_t Database_Revision = 0;
 /* Auto_Slave_Discovery */
 /* Slave_Address_Binding */
 /* Profile_Name */
-static BACNET_REINITIALIZED_STATE Reinitialize_State = BACNET_REINIT_IDLE;
-static const char *Reinit_Password = "filister";
+static BACNET_DEVICE_REINITIALIZE_DATA Reinitialize_Data = {
+    .State = BACNET_REINIT_IDLE, .Password = "filister"
+};
 static write_property_function Device_Write_Property_Store_Callback;
 static list_element_function Device_Add_List_Element_Callback;
 static list_element_function Device_Remove_List_Element_Callback;
 /* backup and restore */
 #if defined BACNET_BACKUP_RESTORE
-/* number of backup files */
-#ifndef BACNET_BACKUP_FILE_COUNT
-#define BACNET_BACKUP_FILE_COUNT 1
-#endif
 
 /* device A will read the Configuration_Files property of the Device object.
    This property will be used to determine the files to read and in what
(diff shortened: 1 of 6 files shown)


skarg (approved):

kim-com:
Let me check it.

kim-com:
@skarg I have one concern before expanding this further.

The routed virtual-device path stores reinitialize/backup state in `DEVICE_OBJECT_DATA`, while the normal non-routed path still uses file-scope state in `device.c`. This creates overlapping concepts in different storage locations, and I’m worried that supporting both paths will keep adding more `#ifdef BAC_ROUTING` / `#ifdef BACNET_BACKUP_RESTORE` branches.

For this PR, is the current limited per-device state approach acceptable, or would you recommend a cleaner follow-up abstraction, possibly aligned with `feature/refactor-device-object-handling` from #1287?

kim-com:
Additional note: I think we may also need to consider future additions for other per-device items, not only the current reinitialize/backup state. Looking forward to your guidance and suggestions.

like this:
- `System_Status`
- `Vendor_Name`,
-  `Model_Name`
- `Application_Software_Version`
- `Firmware_Revision`
- `Location`
- `Serial_Number`
- `Device_UUID`

kim-com:
@skarg I addressed the Copilot feedback above and did a small refactoring pass in the latest push.

I'd like your opinion before taking this further. There are still some device-specific items we may need to add later, so I want to make sure the Backup/Restore compile-time contract is correct.

When `BACNET_BACKUP_RESTORE` is defined, is it valid to treat `BACNET_BACKUP_FILE_COUNT` as always having at least one entry (`BACNET_BACKUP_FILE_COUNT >= 1`, i.e. `> 0`)? The current Backup/Restore path uses `Configuration_Files[0]`, so a zero-file configuration seems to expose only state/property behavior without a usable backup file. I enforced that invariant and removed the `BACNET_BACKUP_FILE_COUNT == 0` handling from the active Backup/Restore path.

skarg:
For this PR the current limited per-device state approach acceptable.  I think it would be better - when someone has time to do it - to have a cleaner follow-up abstraction possibly aligned with feature/refactor-device-object-handling from #1287 as a lot of developers use their own device object and objects and the common handling is often missed, especially as our BACnet committee and BTL creates more rules and adds nuances that require updates in a device object implementation.

Your understanding of BACNET_BACKUP_RESTORE is how I would have done it. Thanks for the contribution!

PR #1378, opened 2026-06-08
hostnport header: fix C++ compatibility

anthony-loiseau-act:
C has no namespaces but C++ has.
Declaring a struct inside a struct makes it globally available in C, but only locally available with a C++ compiler.

Move two nested struct definitions to top-level since those are used later outside the nest.


skarg (approved):

skarg:
Thank you for the fix!
Although I didn't know about the C++ incompatiblilty, I had refactored those structures out of the structure in a WIP branch:
https://github.com/bacnet-stack/bacnet-stack/compare/master...feature/bbmd-host-lookup-table

PR #1426, opened 2026-07-16
Alert enrollment

ryanjmulder:
Added basic Alert Enrollment object.
Alerts are queued into a ring buffer using `Alert_Enrollment_Queue_Alert` and then sent via Notification Class during the `Alert_Enrollment_Intrinsic_Reporting` hook


skarg (approved):

ryanjmulder:
@skarg Hi Steve, sorry for the delay here, I've now addressed all of copilot's review comments

PR #1432, opened 2026-07-20
fix: bounds check missing when parsing Weekly_Schedule day entries from ASCII

94xhn:
`bacapp_parse_application_data()` builds a `BACnetWeeklySchedule` value out of an ASCII string (this is what `bacwp` and similar demo apps use to turn a `--value` argument into something that can go out in a WriteProperty request). For each day it walks the comma-separated time/value tokens and writes them straight into `dsch->Time_Values[tvnum]`, but nothing checks `tvnum` against the array size (`BACNET_DAILY_SCHEDULE_TIME_VALUES_SIZE`, 40) before writing. If a day's brackets contain more than 40 entries, the loop just keeps going and writes past the end of that day's `Time_Values[]` into whatever sits next in memory - the following day's schedule inside `weeklySchedule[]`, and beyond that if you feed it enough entries.

I checked this against the real parser rather than a synthetic reproduction. Feeding a schedule with, say, 60 time-value pairs on Monday and empty brackets for the rest of the week corrupts the following days - Tuesday's `Time_Values[0]` ends up holding data that only Monday's input produced, even though the string said `Tue: []` - and the function still reports success. With a few hundred entries in one day it runs past the local `BACNET_APPLICATION_DATA_VALUE` entirely and crashes.

The sibling parser for `BACnetSpecialEvent` day-schedules (`daily_schedule_from_ascii`, a bit further down in the same file) already guards against exactly this - it looks like the check was simply missed when `parse_weeklyschedule()` was written. This patch adds the same guard: once a day already has `BACNET_DAILY_SCHEDULE_TIME_VALUES_SIZE` entries, reject the string instead of writing past the array.

Added a test next to the existing `test_parse_weeklyschedule` cases that feeds 41 entries for one day and checks both that parsing is now rejected and that the next day's data is left untouched. Ran the full `test_bacapp` suite before and after: without the fix the new case fails (parser reports success on the oversized input); with the fix it passes and nothing else in the suite changed.

One clarification since Weekly_Schedule shows up a lot in past advisories here: this is the ASCII/CLI-side parser, not the wire decoder. `bacnet_dailyschedule_context_decode()` in dailyschedule.c already passes `ARRAY_SIZE()` through correctly, so this isn't reachable by decoding a WriteProperty request off the network - it only affects code that builds a Weekly_Schedule value from a string through this API.

### DIFF
--- a/src/bacnet/bacapp.c
+++ b/src/bacnet/bacapp.c
@@ -4391,6 +4391,11 @@ parse_weeklyschedule(const char *str, BACNET_APPLICATION_DATA_VALUE *value)
                 *sp = '\0';
                 v = bacnet_ltrim(sp + 1, " ");
 
+                if (tvnum >= BACNET_DAILY_SCHEDULE_TIME_VALUES_SIZE) {
+                    /* too many time-value pairs for one day */
+                    return false;
+                }
+
                 /* Parse time directly into the Time_Value slot */
                 if (!datetime_time_init_ascii(
                         &dsch->Time_Values[tvnum].Time, t)) {
--- a/test/bacnet/bacapp/src/main.c
+++ b/test/bacnet/bacapp/src/main.c
@@ -1739,11 +1739,13 @@ static void test_parse_weeklyschedule(void)
 
         for (i = 0; i < 41; i++) {
             snprintf(
-                pair, sizeof(pair), "%s%02d:%02d:%02d.00 TRUE", (i == 0) ? "" : ", ",
-                (i / 3600) % 24, (i / 60) % 60, i % 60);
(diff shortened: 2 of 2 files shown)


skarg (approved):

skarg:
running "pre-commit run --all-files" should clear up the pipeline failure, after another commit of the style fixes.

94xhn:
> running "pre-commit run --all-files" should clear up the pipeline failure, after another commit of the style fixes.

Thanks. I checked the failed CI log, applied the exact clang-format changes to `test/bacnet/bacapp/src/main.c`, and pushed commit `08f0809b4`. The selective clang-format pre-commit hook passes locally; the functional checks were already green on the prior revision.

PR #1481, opened 2026-08-26
Append PROP_PROPERTY_LIST to RPM ALL requests

anthony-loiseau-act:
`Propery_List` property is special, it is voluntary not part of required/optional/proprietary properties lists but it still exists since protocol 14.

Automatically return it for special `Read_property_Multiple ALL` requests (still not returning it in all other RPM requests).

Do it by altering both RPM_Object_Property and RPM_Object_Property_Count

### DIFF
--- a/src/bacnet/basic/service/h_rpm.c
+++ b/src/bacnet/basic/service/h_rpm.c
@@ -44,6 +44,9 @@ static uint8_t Temp_Buf[MAX_APDU] = { 0 };
  * to fetch.
  * @param index The index of the property to fetch.
  * @return The property ID or -1 if not found.
+ * @note For PROP_ALL with protocol revision >= 14, PROP_PROPERTY_LIST is
+ * appended as the last entry, since it is intentionally absent from the
+ * Required/Optional/Proprietary lists.
  */
 static BACNET_PROPERTY_ID RPM_Object_Property(
     struct special_property_list_t *pPropertyList,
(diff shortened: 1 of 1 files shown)


skarg:
Is there an exception that I'm not aware of?

**ANSI/ASHRAE Standard 135.1-2023**
**9. APPLICATION SERVICE EXECUTION TESTS**
**9.20.1.7 Reading ALL Properties**
Purpose: To verify the ability to correctly execute a ReadPropertyMultiple service request that uses the special property
identifier ALL. One instance of each object-type supported is tested.
Notes to Tester: If a property which is not readable using the ReadPropertyMultiple service is in the specified object, and
Protocol_Revision < 7, then either no entry is returned, or an error code is returned. If Protocol_Revision >= 7, then the
entry shall contain 'Error Class': PROPERTY and 'Error-Code': READ_ACCESS_DENIED for that property. _Property_List
(371) shall not appear in the List of Results._

skarg:
**ANSI/ASHRAE Standard 135-2024**
**15. OBJECT ACCESS SERVICES**
**15.7 ReadPropertyMultiple Service**
**15.7.3.1.2 List of Property References**

This parameter shall be a list of one or more BACnetPropertyReferences, each of which corresponds directly to a specific property of the object identified above. The property identifier ALL means that all defined properties of the object are to be accessed, including any proprietary properties.

The property identifier REQUIRED means that only those standard properties having a conformance code of "R" or "W" shall be returned. The property identifier OPTIONAL means that only those standard properties present in the object that have a conformance code "O" shall be returned. If no optional properties are present then no entries associated with OPTIONAL property identifier shall be in the 'List of Results' and if no other properties were requested, then the 'List of Results' shall be present and empty. _The Property_List property shall not be returned when properties ALL or REQUIRED are requested._ See the specification for the particular object type in Clause 12. If the property identifier ALL, REQUIRED, or OPTIONAL is specified and any of the selected properties is not readable by this service, then a Property Access Error for that property shall be returned in the List of Read Access Results as specified by Clause 15.7.3.2.

anthony-loiseau-act:
Sorry, I have not the standard and seen hare and there on the Internet that Property_List can or should be returned in RPM ALL calls (without evidences of this). Reading your lines it appears clear that this is wrong. I have no evidences on this point and you likely have some :)

PR #1484, opened 2026-08-28
CSV: Changed flag should not be cleared upon write

anthony-loiseau-act:
A remote peer must not be able to clear Changed flag of a CSV by writing current value to it, since this would break COV if COV handler is not executed upon every single recv loop


skarg:
Thank you for finding and patching the problem!

Using an 'if' expression would be much clearer instead of the '|=':
```
    if (!characterstring_same(&pObject->Present_Value, present_value)) {
        pObject->Changed = true;
    }            
```

anthony-loiseau-act:
sure, modified

skarg (approved):

PR #1495, opened 2026-09-04
fix(gateway2): Do not announce virtual network as real network

anthony-loiseau-act:
Gateway2 app creates a virtual network with virtual devices on it.

It announces itself as a router to that virtual network. I_Am_Router_To_Network anounce is broadcasted onto the local real network. OK.

But it must not announce virtual network number inside a Network_Number_Is message, whose goal is to announce the real network number if known by the router (and likely useless to broadcast on boot).

Remove this line, gateway2 do not even know its localnetwork number.


anthony-loiseau-act:
Note: I added a very small netport doxygen typo fix in this small PR not open a dedicated PR for such a minor thing

skarg (approved):

PR #1499, opened 2026-09-09
fix(Binary Value): change Present_Value_Set() drive

ghislain-mange-act:
Change `Present_Value_Set()` drive from BACnet Priority 16 to `Relinquish_Default_Set()`. Change applies to binary value commandable namespace only.

This implements comment from previous change:
https://github.com/bacnet-stack/bacnet-stack/pull/1490#discussion_r3935125899

### DIFF
--- a/src/bacnet/basic/object/bv.c
+++ b/src/bacnet/basic/object/bv.c
@@ -569,6 +569,13 @@ unsigned Binary_Value_Present_Value_Priority(uint32_t object_instance)
 
 /**
  * @brief For a given object instance-number, sets the present-value
+ *
+ * When the object is commandable, no priority is given here, so the value
+ * is written wherever the present-value is currently resolved from: the
+ * slot of the active command, or relinquish-default when every priority
+ * slot is relinquished. Either way the present-value becomes the value
+ * given, which is what a caller of this function asks for.
+ *
  * @param  object_instance - object-instance number of the object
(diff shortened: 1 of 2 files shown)


src/bacnet/basic/object/bv.c
```diff
@@ -588,11 +588,9 @@ bool Binary_Value_Present_Value_Set(
 {
     bool status = false;
 #if defined(BACNET_OBJECT_BINARY_VALUE_COMMANDABLE)
-    /* with no priority given, drive present-value at the lowest priority
-       instead of through Relinquish_Default_Set(), so that a genuine
-       Present_Value transition raises Change_Of_Value */
-    status = Binary_Value_Present_Value_Priority_Set(
-        object_instance, value, BACNET_MAX_PRIORITY);
+    /* with no priority given, the closest equivalent is to drive
```
skarg: Would the [principle of least surprise](https://principles-wiki.net/principles:principle_of_least_surprise) instead lead us to evaluate the priority array and set the value at the current commanded priority, or relinquish default if no priority slots are in use? In both cases this function would 'set the present-value' as expected.

ghislain-mange-act: Nice idea - it's been updated accordingly.

ghislain-mange-act:
I also included another commit with tests for future compatibility.

skarg (approved):

